PR #33, opened 2018-09-12
Switches Access Boston to use Yar for session

fionawhim:
Server-side sessions will give us the ability to invalidate the session
without relying on cookie timeouts or clearing.

Adds an auth scheme to hapi-common for common browser-based
authentication with a redirect to a login form.

Also:
 - Does some Hapi updates to keep everything on the same page
 - Removes "engine" from 311-indexer since the latest Node container no
   longer matches it


services-js/access-boston/src/server/access-boston.ts
```diff
@@ -90,8 +90,32 @@ export async function makeServer(port) {
...
+    plugin: yar,
+    options: {
+      // Always stores everything in the cache, so we can clear out sessions
+      // unilaterally rather than relying on cookie expiration and being
+      // vulnerable to replays.
+      maxCookieSize: 0,
+      cookieOptions: {
+        password:
+          process.env.SESSION_COOKIE_PASSWORD ||
+          'iWIMwE69HJj9GQcHfCiu2TVyZoVxvYoU',
```
jessicamarcus: Do we want this dummy string to go into production?

fionawhim: Yeah, it's fine, so we can run w/o any configuration. I can change it to be like "test-xxxx" or "fake-xxxx" so it's obvious that it's on purpose?

There's code that enforces that in prod the password is set in the environment.

jessicamarcus:
looks good to me!

jessicamarcus (approved):

PR #34, opened 2018-09-12
Keeps the dead from aging

fionawhim:
Fixes age calculation to use date of death rather than current date.

Also restores devDependencies that are called directly from the
package.json strings.


jessicamarcus (approved):
lgtm! 👻

PR #55, opened 2018-09-21
Removes styled-jsx from registry-certs

fionawhim:
When doing compilation for Storybook the <style jsx> elements aren't
being processed by the styled-jsx plugin, at least not before TypeScript
converts them.

Since we're not using styled-jsx for new stuff anyway, better to just
convert this than spend too much time trying to fix it.


fionawhim:
There will be a handful of Percy changes from things that got broken by the Babel 7 update. This should be setting them right.

jessicamarcus (approved):
lgtm! I'd forgotten how ugly styled-jsx is... 😆

fionawhim:
Missed two bits in the receipt page, pushing an update.

PR #58, opened 2018-09-24
Start of Access Boston registration flow

fionawhim:
- Adds new flags to session for whether the user needs a new password
   or needs MFA signup
 - Adds those flags to the GraphQL Account type
 - Creates a starter "Register" page that links to change password
 - Modifies Change Password to have a "new password" mode with different
   titles and explanatory text

Does not implement any aspect of MFA registration right now.

### DIFF
--- a/services-js/access-boston/src/pages/change-password.tsx
+++ b/services-js/access-boston/src/pages/change-password.tsx
@@ -59,13 +59,20 @@ interface FormValues {
   crumb: string;
 }
 
+function successUrl(account): string {
+  return account.needsMfaDevice
+    ? `/mfa?message=${FlashMessage.CHANGE_PASSWORD_SUCCESS}`
+    : `/?message=${FlashMessage.CHANGE_PASSWORD_SUCCESS}`;
+}
+
 export default class ChangePasswordPage extends React.Component<Props, State> {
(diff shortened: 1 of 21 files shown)


fionawhim:
This also changes the font for “New Passwords Must:” to reduce the number of different font styles / sizes on the page.

services-js/access-boston/src/pages/change-password.tsx
```diff
@@ -176,15 +197,35 @@ export default class ChangePasswordPage extends React.Component<Props, State> {
...
+
+                  {setNewPassword && (
+                    <>
+                      <p className="t--info m-t500 m-v300">
+                        You’ll need to create a new, secure password to set up
+                        your Access Boston account.
+                      </p>
+
+                      <p className="t--info m-v300">
+                        This password will be good for a{' '}
```
jessicamarcus, line 214: Why do we need to explicitly add a space here?

jessicamarcus (approved):

services-js/access-boston/src/pages/change-password.tsx
fionawhim: JSX collapses whitespace before HTML tags.

PR #59, opened 2018-09-24
Form work, reorganization

jessicamarcus:
This commit represents in-progress work, as I don't want this branch to drift away from `develop` too far.

Layout for the application form has been adjusted, components extracted and/or reorganized, and "prefix" has been entirely removed after discussion with @paylakatel.


modules-js/react-fleet/src/form-elements/SelectDropdown.stories.tsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+import SelectDropdown from './SelectDropdown';
+
+storiesOf('Form Elements/Select', module)
+  .add('Default Dropdown', () => (
+    <React.Fragment>
```
fionawhim: You can also use the shorthand `<>` instead of `<React.Fragment>`

modules-js/react-fleet/src/form-elements/SelectDropdown.tsx
```diff
@@ -0,0 +1,81 @@
+import React, { SelectHTMLAttributes } from 'react';
+import hash from 'string-hash';
+
+
+interface Props extends SelectHTMLAttributes<HTMLSelectElement> {
+  label: string;
+  error?: string | boolean;
+
+  variant?: 'small';
```
fionawhim: I’d recommend making `small` its own optional boolean prop. That lets you mix variants more easily.

modules-js/react-fleet/src/form-elements/SelectDropdown.tsx
```diff
@@ -0,0 +1,81 @@
+import React, { SelectHTMLAttributes } from 'react';
+import hash from 'string-hash';
+
+
+interface Props extends SelectHTMLAttributes<HTMLSelectElement> {
```
fionawhim: I think if you’re going to use `SelectHTMLAttributes` then you need to pass all unused `props` to the `<select>` element.

If you want to only have some props on `<select>`, then enumerate the explicitly instead of `extends`ing.

modules-js/react-fleet/src/form-elements/SelectDropdown.tsx
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * Automatically generates an id for the element if one is not provided.
+ *
+ * If hideLabel is specified, the label text will be added as an aria-label
+ * attribute on the select element.
+ *
+ * todo: at narrow widths, padding visually obscures current selection
+ *
+ * todo: currently assumes it will receive an array of strings to create its <option>s; there may be cases when we want an option to have a different value from its displayed text, and pass in an array of objects instead. - 9/21 jm
+ */
```
fionawhim, line 39: Good comment.

modules-js/react-fleet/src/form-elements/SelectDropdown.tsx
```diff
@@ -0,0 +1,81 @@
...
+export default function SelectDropdown(props: Props): JSX.Element {
+  const { error, hideLabel, label, required, variant } = props;
+  const id = props.id || `select-${hash(label)}`;
+
+  const classNames = {
```
fionawhim: Interesting pattern. It’s nice for documentation, but on the other hand it doesn’t give you an unused variable warning if you stop using one of them.

modules-js/react-fleet/src/form-elements/SelectDropdown.tsx
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        {required && <span className="t--req">Required</span>}
+      </label>
+
+      <div className={classNames.container}>
+        <select
+          id={id}
+          className={classNames.select}
+          aria-label={hideLabel ? label : ''}
```
fionawhim: 👍

modules-js/react-fleet/src/form-elements/SelectDropdown.tsx
```diff
@@ -0,0 +1,81 @@
...
+          required={required}
+
+          onBlur={props.onBlur}
+          onChange={props.onChange}
+          onFocus={props.onFocus}
+        >
+          {!props.hideBlank && <option />}
+
+          {props.options.map(option => (
+            <option key={`${id}_${option}`}>{option}</option>
```
fionawhim: The `id` is not strictly necessary in here, since `key` doesn’t have to be globally unique.

modules-js/react-fleet/src/form-elements/inputs/TextInput.tsx
```diff
@@ -0,0 +1,88 @@
...
+ * class if there's an error.
+ *
+ * If hideLabel is specified, the label text will be added as an aria-label
+ * attribute on the input element.
+ */
+interface Props extends InputHTMLAttributes<HTMLInputElement> {
+  label: string;
+  error?: string | boolean;
+
+  variant?: 'small';
```
fionawhim: same note about making this `small?: boolean`

modules-js/react-fleet/src/form-elements/inputs/TextInput.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+        required={required}
+        inputMode={props.inputMode}
+        maxLength={props.maxLength}
+        minLength={props.minLength}
+
+
+        onBlur={props.onBlur}
+        onChange={props.onChange}
+        onFocus={props.onFocus}
```
fionawhim, line 90: I’m curious why you went the enumeration route for all these `<input>` props.

modules-js/react-fleet/src/layouts/AppLayout.stories.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { storiesOf } from '@storybook/react';
 
 import AppLayout from './AppLayout';
-import SectionHeader from '../elements/SectionHeader';
+import SectionHeader from '../sectioning-elements/SectionHeader';
```
fionawhim, line 5: Descriptive category. I like it.

modules-js/react-fleet/src/sectioning-elements/SectionHeader.tsx
```diff
@@ -0,0 +1,34 @@
...
+};
+
+export default function SectionHeader(props: Props): JSX.Element {
+  const { title, yellow, subheader } = props;
+  const classes = [
+    'sh',
+    'm-b300'
+  ];
+
+  if (yellow!) {
```
fionawhim: `!` I think shouldn’t be necessary here; I would expect `undefined` is a valid statement in an `if`.

services-js/commissions-app/src/client/ApplicantInformationSection.tsx
```diff
@@ -0,0 +1,175 @@
...
+import { Field } from 'formik';
+import { css } from 'emotion';
+
+import { MEDIA_MEDIUM } from '@cityofboston/react-fleet';
+
+import TextInputContainer from './common/TextInputContainer';
+
+
+const COMMON_MARGIN = 0.75;
+
```
fionawhim: These CSS chunks could probably use an overall comment explaining the strategy. You might also want a name a bit more specific than `COMMON`. Does `ROW_STYLING` fit?

fionawhim: You should include a note about why you went w/ this and not the grid classes.

services-js/commissions-app/src/client/ApplicantInformationSection.tsx
```diff
@@ -0,0 +1,175 @@
...
+            label="Zip"
+            name="zip"
+
+            placeholder="Zip Code"
+            required
+          />
+        </div>
+      </div>
+
+      <Field
```
fionawhim, line 154: In a future change, think about how to make this a `tel` input field.

services-js/commissions-app/src/client/ApplicantInformationSection.tsx
```diff
@@ -0,0 +1,175 @@
...
+        component={TextInputContainer}
+        label="Phone"
+        name="phone"
+
+        placeholder="Phone Number"
+        required
+      />
+
+      <Field
+        component={TextInputContainer}
```
fionawhim, line 164: Same with this and an `email` type.

yarn.lock
```diff
@@ -49,6 +70,26 @@
     semver "^5.4.1"
     source-map "^0.5.0"
 
+"@babel/core@7.0.0-beta.42":
```
fionawhim: ruh roh. something snuck back in.

fionawhim (changes requested):

modules-js/react-fleet/src/form-elements/SelectDropdown.stories.tsx
jessicamarcus: I generally opt to be explicit, but I'm happy to use the shorthand in our codebase.

modules-js/react-fleet/src/form-elements/SelectDropdown.tsx
jessicamarcus: ah, good point.

yarn.lock
jessicamarcus: 😭

fionawhim (approved):
LGTM!

PR #61, opened 2018-09-25
small styling adjustments

jessicamarcus:
When I pushed to staging, I realized there were a few small issues... Some styling was tweaked, and I removed the incorrect "required" attribute from phone number field.


fionawhim (approved):
LGTM!

PR #63, opened 2018-09-25
PingID backend support for MFA registration

fionawhim:
Adds API for MFA registration to Access Boston:

 - Adds a PingId service to communicate with PingID
 - Pulls name and email information from SAML logins and stores it in
   the session
 - Adds GraphQL mutations for starting and finishing MFA registration

Also:
 - Cleans up SamlAuthFake to use promisify
 - Moves password mutations out of schema.ts and into workflows.ts
   because it was getting big


jessicamarcus (approved):
I haven't used PingID before, but this looks perfectly reasonable to me. :)

fionawhim:
That makes two of us. :)

Thanks for the review.

PR #69, opened 2018-09-28
Fixes when SAML assertion has empty attributes

fionawhim:
For forgot password, there’s no first name, last name, or email address.

Also adds hapi-dev-errors so that when errors come up we see a shiny
error page / stack trace rather than the default 500 JSON response.


jessicamarcus (approved):
lgtm!

PR #70, opened 2018-10-01
Adds UI for device registration flow

fionawhim:
- mfa.tsx page to manage the verification form and API calls
 - DeviceVerificationForm stateless component to show form w/ choice
   between phone and email, and SMS and voice
 - DeviceVerificationModal box for accepting the 6-digit code
 - New validation to make sure that secondary email addresses are non-City

Also:
 - Factors out StatusModal from the various forms
 - Moves page model classes into subdirectory of integration tests
 - Changes Radio group to not have bottom padding by default, adds it to
   FacetList instead
 - Fixes formatting in react-fleet input controls with Prettier
 - Changes generated GraphQL TypeScript types to a .ts file so it can
   export enums
 - Adds a render func to the Access Boston TextInput class to allow for
   side-by-side error messages and submit buttons
 - APP_RUNNING state added to window so that TestCafe can know that
   the page is ready to interact with


services-js/access-boston/src/client/TextInput.tsx
```diff
@@ -19,18 +37,33 @@ interface Props extends InputHTMLAttributes<HTMLInputElement> {
  *
  * All props besides "label" and "error" are passed along as props for the
  * <input> element.
+ *
+ * TODO(finh): Merge this with the react-fleet TextInput.
```
jessicamarcus, line 41: When we get to this point let's talk through theory and come up with consensus, since I went with a different approach with the current react-fleet version.

services-js/access-boston/src/integration/change-password.testcafe.ts
```diff
@@ -1,7 +1,7 @@
 import { Selector } from 'testcafe';
```
jessicamarcus, line 1: Thanks for the reminder to chat with you about the ifs and whens on bringing testcafe in for other apps!

jessicamarcus (approved):

services-js/access-boston/src/client/TextInput.tsx
fionawhim: 👍

services-js/access-boston/src/integration/change-password.testcafe.ts
fionawhim: Yeah, lemme know any time!

PR #71, opened 2018-10-01
Code verification part of Access Boston register

fionawhim:
- Adds a new mutation query for the frontend to verify the code
 - Calls the new query from the modal
 - Adds a "done" landing page that clears out the session
 - Generalizes form auto-submission into a "RedirectForm" component
 - Goes to "done" after change password if there's no MFA to do

Also:
 - Removes a change password test that got flaky, since we're going to
   delete non-JS soon anyway


services-js/access-boston/src/client/RedirectForm.tsx
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * You can also provide children that will be rendered in the form. This is
+ * useful for putting in a submit button.
+ */
+export default class RedirectForm extends React.Component<Props> {
+  private readonly formRef = React.createRef<HTMLFormElement>();
+
+  redirect() {
+    const form = this.formRef.current;
+    if (!form) {
```
jessicamarcus: This could be refactored to `if (form) form.submit();`

services-js/access-boston/src/client/device-verification/DeviceVerificationModal.tsx
```diff
@@ -121,6 +127,33 @@ export default class DeviceVerificationModal extends React.Component<Props> {
...
+
+    return (
+      <>
+        <div className="t--intro">
+          <span className="t--err">Something went wrong.</span>
+        </div>
+
+        <div className="m-v400">
+          We had a problem verifying that code. You can try to{' '}
+          <a href="javascript:void(0)" onClick={resendVerification}>
```
jessicamarcus: `<button type="button" onClick={resendVerification}>` would be more semantic/accessible here (and on line 145).

services-js/access-boston/src/client/RedirectForm.tsx
fionawhim: Done.

services-js/access-boston/src/client/device-verification/DeviceVerificationModal.tsx
fionawhim: Updated. Requires CityOfBoston/patterns#380 for new styles.

jessicamarcus (approved):
lgtm!

jessicamarcus:
Actually, after reviewing the percy screenshots, we probably want to change the "resend code" buttons' styling to appear unchanged from when they were `<a>` links.

fionawhim:
Yeah, just waiting on the patterns change to go through. Then I'll re-run Percy on this PR.

jessicamarcus:
Ah, right - I didn't notice that that build was failing. nm! 🙃

PR #73, opened 2018-10-02
Removes the non-JS password workflows

fionawhim:
Ping requires JavaScript, so there's not much value in us working
without it. Removing this speeds up tests and reduces the number of
things that could break.


jessicamarcus (approved):
End of a era! Looks good to me.

PR #76, opened 2018-10-03
Language/style tweaks for Access Boston register

fionawhim:
- Updates some language to match the document from IAM team. All still
   a work-in-progress.
 - Switches "Change Password" style to use a squiggly <hr/> rather than
   a gray block to separate text and form.


jessicamarcus (approved):
lgtm!

PR #78, opened 2018-10-03
Converts registry-certs to use common Hapi/Next

fionawhim:
Changes registry-certs to use hapi-next for its routes. Adds the
compression filter and custom route support to these functions.

Creates routes in hapi-common for serving static assets and storybook,
based on the ones from registry-certs.

Switches registry-certs to use fetchGraphql instead of its custom
loopbackGraphql. Since fetchGraphql gets its inject function from the
config, we can remove RequestAdditions.

Also:

 - Removes stripTrailingSlashes because it interferes with redirecting
   /storybook -> /storybook/, which is necessary for assets to load
   relatively.
 - Adds web api key and Stripe key to the Next config instead of
   __NEXT_DATA__
 - Adds static Storybook and favicon to access-boston service.
 - Removes legacy "req.finished = true" for redirects. It’s not needed
   anymore.


fionawhim:
ok this PR kinda got away from me. But it deletes more than it adds, so that’s good.

modules-js/hapi-next/src/hapi-next.ts
```diff
@@ -82,22 +100,49 @@ export function makeRoutesForNextApp(
...
+): Lifecycle.Method {
+  return async (request: HapiRequest, h: ResponseToolkit) => {
     const {
       raw: { req, res },
+      query,
+      path,
+      params,
     } = request;
 
+    // Be consistent about having Next pages not have a slash. Unless it’s only
```
jessicamarcus: Minor, but adds clarity:

`// Be consistent about having Next pages not have a` **`trailing`** `slash. Unless it’s only`

services-js/registry-certs/client/dao/CheckoutDao.test.ts
```diff
@@ -1,6 +1,5 @@
-// @flow
+import { FetchGraphql } from '@cityofboston/next-client-common';
```
jessicamarcus, line 1: I know this change affected a lot of files, but I think this name change makes things much more clear!

jessicamarcus (approved):

modules-js/hapi-next/src/hapi-next.ts
fionawhim: 👍

services-js/registry-certs/client/dao/CheckoutDao.test.ts
fionawhim: Yay!

fionawhim:
Waiting for tests to pass again before merging.

Fixed an issue in registry-certs' storybook.

PR #83, opened 2018-10-05
Updates IdentityIQ on device registration

fionawhim:
Adds the email address and phone number to the session so that we can
send them to IIQ.


jessicamarcus (approved):
lgtm!

PR #86, opened 2018-10-06
updated library component to accept declared styles; added missing class

jessicamarcus:


fionawhim (approved):
LGTM!

And it looks like Prettier is working now on the commissions-search stuff, which is good.

PR #95, opened 2018-10-11
Combines the RegistryData and RegistryOrders files

fionawhim:
These classes don't need to be separate, so we're combining them to
avoid confusion.

Also:
 - Standardizes on Node 8.0.0 TypeScript types. This required some
   changes now that process.env.* is string | undefined, rather than
   any. (Same with querystring, it appears.)


jessicamarcus (approved):
lgtm! Sorry for not reviewing sooner; I'd closed my browser email tab, oops.

PR #96, opened 2018-10-11
Coverts registry-certs to use an App component

fionawhim:
Changes the dependency injection strategy to use an App component
rather than default props. This improves type checking and testing by
making it explicit in Props what dependencies a component has.

Extends the Access Boston pattern slightly by making the GetInitialProps
type have type parameters for which fields the function is interested
in.

Changes Nav and AppLayout to take Cart explicitly, rather than have it
magically come in.

Storyshots changes are coming from using the same Cart object between
the nav bar and the rendered page. (Previously the nav bar used a
default, empty cart in these cases.)


jessicamarcus (approved):
lgtm!

PR #99, opened 2018-10-12
Upgrades Next to 7.0.2

fionawhim:
Fixes security vulnerability.

Also upgrades Storybook so that we're all on the same version for
Webpack (other than create-react-app, which is still on 3).


jessicamarcus (approved):
💯

PR #100, opened 2018-10-12
More death certs naming updates

fionawhim:
- Adds "DeathCertificates" to search and lookup in RegistryDb
 - Adds a parameter to add order / item functions to specify the type of
   certificate
 - Parameterizes cost calculations on certificate type

Also:
 - Separates out RegistryDbFake into its own file, uses it more in tests
   to add type safety around mock methods.


jessicamarcus (approved):
lgtm 👍

PR #106, opened 2018-10-16
Fixes regression in handling of NextJS HMR routes

fionawhim:
We used to be able to pass the URL separately, but the latest version of
Next looks at the request’s "url" property for a match to
on-demand-entries-ping.


jessicamarcus (approved):
🎉

fionawhim:
Regression seemed to have happened from 7.0.0 -> 7.0.2 but I'm not quite sure how. 🤷‍♀️

PR #108, opened 2018-10-17
Prep for apply form submissions

fionawhim:
Moves things around so that form validation can be done on both client-
and server- side. Also changes to support POSTing a FormData object with
fetch so that we can get file uploads.

 - Moves form values interface into validationSchema.ts
 - Changes selection to be by commission ID, not commission object
 - Switches to HTML <form> element so we can get a ref for it for
   creating a FormData
 - Adds "name" attribute to checkboxes and other fields so that they
   show up in FormData
 - Removes touch / blur handling from checkboxes because it seems to
   break FieldArray by overwriting its array value with a boolean
 - Adds all fields to the application form schema and uses the
   ApplyFormValues interface to strongly type it

Also:
 - Tweaks apply page to put <div class="mn"> inside the <AppLayout> and
   moves the stylesheet to _document
 - Adds a <title> element to the form page
 - Changes the ApplyPage stories to exercise pre-checking a commission

### DIFF
--- a/services-js/commissions-app/src/lib/validationSchema.ts
+++ b/services-js/commissions-app/src/lib/validationSchema.ts
@@ -3,18 +3,39 @@ import 'core-js/fn/array/from';
 
 import * as Yup from 'yup';
 
-export const applicationForm = Yup.object().shape({
+export interface ApplyFormValues {
+  firstName: string;
+  middleName: string;
+  lastName: string;
+  streetAddress: string;
+  unit: string;
+  city: string;
+  state: string;
+  zip: string;
+  phone: string;
+  email: string;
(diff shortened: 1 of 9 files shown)


services-js/commissions-app/src/lib/validationSchema.ts
```diff
@@ -3,30 +3,62 @@ import 'core-js/fn/array/from';
 
 import * as Yup from 'yup';
 
-export const applicationForm = Yup.object().shape({
+export interface ApplyFormValues {
```
jessicamarcus, line 6: Good call on moving this interface here.

jessicamarcus (approved):
Thanks!

PR #109, opened 2018-10-17
added polyfill; minor component rework

jessicamarcus:
- forgot to include polyfill for Array.find()
- adjustments to CollapsibleSection component as per @finneganh‘s suggestion.


services-js/commissions-app/src/pages/commissions/apply.tsx
```diff
@@ -1,3 +1,6 @@
+// polyfill Array.find() for IE
+import 'core-js/fn/array/find';
```
fionawhim, line 2: good catch. Thanks!

fionawhim (approved):

PR #110, opened 2018-10-17
Sends submitted application to the database

fionawhim:
- Adds an apply method to CommissionsDao to call the SaveApplication
   stored procedure.

Also:
 - Adds in max length validation that matches the DB columns
 - Hides the error message for not enough commissions checked
 - Fixes auto-checking not working due to incorrect "query"
   destructuring in getInitialProps

### DIFF
--- a/services-js/commissions-app/src/client/CommissionsListSection.tsx
+++ b/services-js/commissions-app/src/client/CommissionsListSection.tsx
@@ -1,5 +1,8 @@
-// polyfill Array.includes() for IE
+// polyfill Array methods for IE
+//
+// TODO(finh): Can we do the auto-import for these?
 import 'core-js/fn/array/includes';
+import 'core-js/fn/array/find';
 
 import React from 'react';
 import { FieldArray, FormikErrors } from 'formik';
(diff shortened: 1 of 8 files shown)


services-js/commissions-app/src/client/CommissionsListSection.tsx
```diff
@@ -99,7 +110,7 @@ export default class CommissionsListSection extends React.Component<Props> {
           />
         </CollapsibleSection>
 
-        {commissionsSelectionError(this.props.errors)}
+        {commissionsSelectionError(errors, selectedCommissionIds.length > 0)}
```
jessicamarcus, line 115: shouldn’t this be `selectedCommissionIds.length === 0`?

services-js/commissions-app/src/server/commissions-app.ts
```diff
@@ -163,18 +164,23 @@ export async function makeServer(port, rollbar: Rollbar) {
   server.route({
     method: 'POST',
     path: '/commissions/submit',
+    options: {
+      payload: {
+        // Two 5mb files in bytes, plus another 500k for other form information.
```
jessicamarcus, line 169: If we’re changing the max accepted value, we’ll also need to change the helper text, as well as the `sizeLimit` property on the two `FileInput`s.

https://github.com/CityOfBoston/digital/blob/18f5d780b54e9f6613f76c8ca05cd2acf6727519/services-js/commissions-app/src/client/ApplicationForm.tsx#L152
https://github.com/CityOfBoston/digital/blob/18f5d780b54e9f6613f76c8ca05cd2acf6727519/services-js/commissions-app/src/client/ApplicationForm.tsx#L158

jessicamarcus (changes requested):
Other than the file upload size limit changes, lgtm.

services-js/commissions-app/src/client/CommissionsListSection.tsx
fionawhim: No, we want to hide the error when length is 0, not show it. The idea is that this isn't a useful error to show, especially because we don't currently have blur handling.

services-js/commissions-app/src/server/commissions-app.ts
fionawhim: Good catch. Updated in those two places.

jessicamarcus (approved):

PR #111, opened 2018-10-18
Storybook / form cleanup for Commissions App

fionawhim:
- Moves Formik to the ApplyPage to make ApplicationForm depend solely
   on its props
 - Adds more stories around the form states and also the submitted state
 - Adds static storybook to deploy
 - Adds error state, close box, and stories to react-fleet’s StatusModal
 - Upgrades Formik to 1.0.x to match access-boston
 - Removes use of Formik’s Field component because it requires context
   (and therefore is not easily renderable via Storybook)
 - Makes resume and cover letter optional again
 - Moves enabling legislation and no open positions text to only appear
   once on the form

Also:
 - Updates hapi-common to not report 404s to Rollbar
 - Adds favicon to commissions-app
 - Tweaked error type for Textarea to match Formik
 - Sends form JS errors to Rollbar


services-js/commissions-app/src/pages/commissions/apply.tsx
```diff
@@ -5,20 +5,22 @@ import fetchCommissions, {
...
 
 interface Props {
   commissions: Commission[];
   commissionID?: string;
+  testSubmittedPage?: boolean;
 }
 
 interface State {
-  isSubmitting: boolean;
   applicationSubmitted: boolean;
```
jessicamarcus, line 24: `isSubmitting` is intended to be used to provide visual feedback to the user while the server is processing their application; I know it's not in the specs, but it might be nice to at least change the cursor to `wait`, and maybe change the submit button text to "submitting...". What do you think?

services-js/commissions-app/src/client/ApplicantInformationSection.tsx
```diff
@@ -70,30 +69,33 @@ const SPLIT_STYLING = css(`
   }
 `);
 
-export default function ApplicantInformationSection(): JSX.Element {
+interface Props {
+  // Slight hackery to avoid passing all the Formik props this deeply. We prefer
```
jessicamarcus, line 73: til - thanks!

jessicamarcus (approved):

services-js/commissions-app/src/pages/commissions/apply.tsx
fionawhim: Changed button text to "Submitting…". Added wait cursor as an option to `StatusModal`, since that’s appearing while `isSubmitting` is true.

(BTW the `isSubmitting` that was in the `State` for `ApplyPage` is actually now being tracked by Formik. You can see `isSubmitting` as one of the `FormikProps` that `ApplicationForm` requests.)

jessicamarcus: Ah, missed that in `FormikProps` - thx!

PR #112, opened 2018-10-19
passes “required” prop down for native <input> elements

jessicamarcus:
The red ”Required” indicator in the text field labels will appear if the `<input>` element has the `required` attribute present; this change passes it down from the `<Field>` component onto the native element.

 I added the `@ts-ignore` in the Story because Typescript was complaining to me irt the `touched` values. ¯\_(ツ)_/¯


fionawhim (approved):
Can you run `yarn install` and see if the updated version of Formik clears up that TypeScript error?

After that, LGTM. Thanks for catching.

PR #113, opened 2018-10-19
Commissions list design update

fionawhim:
Collapses the list by default if you’ve already selected a commission.
Adds links to the commission’s home page.

Also:
 - Supports the new error state for checkboxes
 - Fixes commissions-app’s static assets to be served at /commissions/


jessicamarcus (approved):
👍

PR #114, opened 2018-10-22
Changes "Field" to a makeField function

fionawhim:
Because "Field" was created in every render() call, React wasn't
reconciling it across renders but instead treaing it as a new component.

This was causing the <input> elements to be completely re-created, which
meant they were losing focus with each keystroke.


jessicamarcus (approved):
Weird... good catch.

PR #115, opened 2018-10-22
Commissions app confirmation emails

jessicamarcus:
This functionality was heavily influenced by that of [registry-certs](https://github.com/CityOfBoston/digital/tree/develop/services-js/registry-certs). It is intended to be called by the `commissions/apply/` handler upon db write success.

There are several outstanding questions/tasks in order to implement:

- A ”newId” value needs to be returned from the server in order to create a hyperlink to the application inside the policy office‘s confirmation email body. In the original code, you can see this in **dal.cs** and in the `EmailNotice()` method in **apply.aspx.cs**.
- The array of selected commissions ids is needed; I’m not certain what the best method to pass these in would be in this case (passed from the FormData? from the server response?)
- The `commissions` array is needed in order to generate the array of applied-for commission name strings for the email bodies; I'm not sure what the best method is to ensure the Email service has access to it (passed from the submission handler? graphQL?)


services-js/commissions-app/src/client/CommissionsListSection.tsx
```diff
@@ -198,6 +198,7 @@ function renderCommission(
             )}
           </div>
         }
+        // @ts-ignore
```
fionawhim: I’d like to avoid `@ts-ignore` in our code. This may be cleared up in latest `develop`, but if not you can tweak the type of Props to fix this.

services-js/commissions-app/src/server/email/EmailContent.stories.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+// We do this from the snapshots generated by running the tests
+// because we don't want to try and run the email templates in the browser.
+//
+// This snapshot parsing is taken from the jest-snapshots utils code. We need to
+// use require rather than fs.readFile so that the snapshot gets webpack-bundled
+// and we can run this client-side.
```
fionawhim, line 9: 👍

services-js/commissions-app/src/server/email/EmailContent.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  };
+
+  const applicantEmailBodies = emailContent.renderApplicantBodies(
+    applicationData
+  );
+  const policyOfficeEmailBodies = emailContent.renderPolicyOfficeBodies(
+    applicationData
+  );
+
+  it('applicant HTML', () => {
```
fionawhim: If you feel like being grammatical, you can change these `it`s to `test`s.

services-js/commissions-app/src/server/email/EmailContent.ts
```diff
@@ -0,0 +1,66 @@
+import fs from 'fs';
+import Handlebars from 'handlebars';
+import mjml2html from 'mjml';
+
+require('./handlebars-helpers');
+
+export interface TemplateData {
+  commissions: Array<string>;
```
fionawhim: Can this be `commissionNames`?

services-js/commissions-app/src/server/email/EmailContent.ts
```diff
@@ -0,0 +1,66 @@
...
+  email: string;
+  newId: string;
```
fionawhim: `applicationId`?

services-js/commissions-app/src/server/email/EmailContent.ts
```diff
@@ -0,0 +1,66 @@
...
+  applicant: readTemplateFiles('applicant'),
+  policyOffice: readTemplateFiles('policyOffice'),
+};
+
+export default class EmailContent {
+  handlebarsApplicantTemplates: any;
+  handlebarsPolicyOfficeTemplates: any;
+
+  constructor() {
+    this.handlebarsApplicantTemplates = readTemplateFiles('applicant');
```
fionawhim: This seems redundant with the `templates` global above.

Between the two, I prefer having all the work happen inside the class when an object is constructed or initialized, rather than at module load time.

services-js/commissions-app/src/server/email/policyOffice.mjml.hbs
```diff
@@ -0,0 +1,44 @@
...
+            <mj-group>
+                <mj-column width="12%" vertical-align="middle">
+                    <mj-image align="left" padding="0" alt="" src="https://patterns.boston.gov/images/public/seal.png"></mj-image>
+                </mj-column>
+
+                <mj-column width="88%" vertical-align="middle">
+                    <mj-text padding="0 0 0 10px" line-height="1.5">
+                        <p>Hello,</p>
+
+                        <p>A new application was received for the following Boards and Commissions:</p>
```
fionawhim: It might be worth putting the applicant name in.

services-js/commissions-app/src/server/email/policyOffice.mjml.hbs
```diff
@@ -0,0 +1,44 @@
...
+
+                        <ul>
+                            {{#each commissions}}
+                                <li>{{this}}</li>
+                            {{/each}}
+                        </ul>
+
+                        <p><a href="http://zpappweb01/cityclerk/commissions/applications/view.aspx?aid={{newId}}">View the application details</a>.</p>
```
fionawhim: This should come from an environment variable so that it’s not checked in.

services-js/commissions-app/src/server/services/Email.ts
```diff
@@ -0,0 +1,117 @@
...
+    this.rollbar = rollbar;
+
+    this.emailContent = new EmailContent();
+
+    // todo: need commissions array
+    this.commissions = [];
+  }
+
+  // todo: need newId from db server
+  sendConfirmations(email, newId, commissionIds) {
```
fionawhim: I'd recommend passing `DbBoard` objects in for the commissions.

services-js/commissions-app/src/server/services/Email.ts
```diff
@@ -0,0 +1,117 @@
...
+            Subject: subject,
+            HtmlBody: renderedBodies.html,
+            TextBody: renderedBodies.text,
+          },
+          (err, result) => (err ? reject(err) : resolve(result))
+        );
+      });
+    } catch (err) {
+      // todo: ? confirm this will work as expected
+      return Promise.reject(err);
```
fionawhim: You don’t need this bit. You actually don’t need to make the function `async`. If an exception is thrown in the `Promise` constructor function, that automatically just rejects the promise. So you can `return new Promise(…)` and that does the right thing.

fionawhim (changes requested):
To answer your questions:

- Yep, we’ll need to get the application ID. Should be feasible once one of us is back on the City Hall network.

- The commissionIds should be available from the `validForm` value in the submit handler.

- For the commission names, I'd recommend having the email functions take an array of DbBoard objects. The submit handler is already creating a `CommissionsDao` object, so you can use that to do the lookup.

(I notice now that `fetchBoard` does not use `DataLoader` to de-dupe. That might be worth doing, or you could just call `fetchBoards` and filter the results.)

services-js/commissions-app/src/server/email/EmailContent.test.ts
jessicamarcus: ...you know me so well! 😄

services-js/commissions-app/src/server/email/EmailContent.ts
jessicamarcus: That was my first impulse, but went back on it because `newId` is what the old code used... if you're thinking the same, then we‘ll happily use `applicationId` instead.

services-js/commissions-app/src/server/email/policyOffice.mjml.hbs
jessicamarcus: @paylakatel what do you think?

paylakatel: That would be great if we can! Could we make the language "A new application was received from _first_ _last_ for the following Boards and Commissions:"

jessicamarcus: Sure can! Would you also like to add “Hello _firstName_,” to the applicant email templates?

paylakatel: Yes! Thank you

services-js/commissions-app/src/server/email/EmailContent.ts
jessicamarcus: That **is** totally redundant - thanks for catching!

services-js/commissions-app/src/server/email/EmailContent.ts
fionawhim: I wouldn't spend any effort trying to match the naming of the old code. Assume that no one will look at it again.

jessicamarcus:
Latest updates bring in email-sending functionality, with an outstanding question: will we need to create new servers on Postmark? If so, I’d like to steal a moment of guidance from you. 😃

services-js/commissions-app/src/lib/validationSchema.ts
```diff
@@ -52,7 +52,10 @@ export const applyFormSchema = Yup.object<ApplyFormValues>({
     .matches(new RegExp(/^\d{5}$/), 'That doesn’t look like a ZIP code.')
     .max(50),
   phone: Yup.string()
-    .matches(PHONE_REGEXP, 'That doesn’t look like a phone number.')
+    .matches(PHONE_REGEXP, {
+      message: 'That doesn’t look like a phone number.',
+      excludeEmptyString: true,
```
fionawhim, line 57: good catch

services-js/commissions-app/src/server/dao/CommissionsDao.ts
```diff
@@ -154,17 +161,25 @@ export default class CommissionsDao {
       .input('degreeAttained', form.degreeAttained || null)
       .input('education', form.otherInformation || null)
       .input('institution', form.educationalInstitution || null)
-      .input('resumeImg', form.resume instanceof Buffer ? form.resume : null)
+      .input(
+        'resumeImg',
+        VarBinaryType, // https://stackoverflow.com/questions/2420708/operand-type-clash-nvarchar-is-incompatible-with-image
```
fionawhim, line 166: ahhh!! Good find. I must have just tested with the files being uploaded, I neglected to test the `null` case.

services-js/commissions-app/src/server/commissions-app.ts
```diff
@@ -177,14 +190,24 @@ export async function makeServer(port, rollbar: Rollbar) {
...
 
       try {
         validForm = await applyFormSchema.validate(req.payload as any);
+        fetchedBoards = await makeCommissionsDao().fetchBoards();
       } catch (e) {
         throw boomify(e, { statusCode: 400 });
       }
 
-      await makeCommissionsDao().apply(validForm);
+      await makeCommissionsDao()
```
fionawhim: Since you’re in an `async` function, you don’t need `then`:

```
const applicationId = await makeCommissionsDao().apply(validForm);
await emailService.sendConfirmations(validForm, fetchedBoards, applicationId);
```

services-js/commissions-app/src/server/email/EmailContent.ts
```diff
@@ -0,0 +1,85 @@
+import fs from 'fs';
+import Handlebars from 'handlebars';
+import mjml2html from 'mjml';
+
+require('dotenv').config();
```
fionawhim: This shouldn’t be necessary. It’s already done by the `start.ts` file.

services-js/commissions-app/src/server/email/EmailContent.ts
```diff
@@ -0,0 +1,85 @@
...
+  html: string;
+  text: string;
+}
+
+export default class EmailContent {
+  handlebarsTemplates: any;
+
+  constructor() {
+    this.handlebarsTemplates = {
+      applicant: '',
```
fionawhim: could you just do:

```
this.handlebarsTemplates = {
  applicant: this.readTemplateFiles('applicant'),
  …
};
```

?

services-js/commissions-app/src/server/email/EmailContent.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  public renderApplicantBodies(data: TemplateData): RenderedEmailBodies {
+    return this.renderEmailBodies('applicant', data);
+  }
+
+  public renderPolicyOfficeBodies(data: TemplateData): RenderedEmailBodies {
+    return this.renderEmailBodies(
+      'policyOffice',
+      data,
+      process.env.COMMISSIONS_URI
```
fionawhim: You may want to pass this in, and keep the `process.env` bits in `commissions-app.ts`

services-js/commissions-app/src/server/email/EmailContent.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  private renderEmailBodies(
+    recipient: 'applicant' | 'policyOffice',
+    data: TemplateData,
+    uri?: string
+  ): RenderedEmailBodies {
+    const templateData = { ...data, uri };
+
+    const htmlTemplate = Handlebars.compile(
```
fionawhim: Ideally you’d `compile` the templates in the `constructor` so that it only has to be done once.

services-js/commissions-app/src/server/services/Email.test.ts
```diff
@@ -0,0 +1,17 @@
+import { namesFromIds } from './Email';
+
+describe('namesFromIds', () => {
```
fionawhim, line 3: Good use of unit test, especially since there’s a type conversion that’s necessary between integer and string.

services-js/commissions-app/src/server/services/Email.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  constructor(from: string, postmarkClient: PostmarkClient, rollbar: Rollbar) {
+    this.from = from;
+    this.postmarkClient = postmarkClient;
+    this.rollbar = rollbar;
+
+    this.emailContent = new EmailContent();
+  }
+
+  // todo: need newId from db server
```
fionawhim: can remove TODO

services-js/commissions-app/src/server/services/Email.ts
```diff
@@ -0,0 +1,109 @@
...
+    const renderedPolicyOfficeBodies = this.emailContent.renderPolicyOfficeBodies(
+      applicationData
+    );
+
+    // send to applicant...
+    this.sendConfirmationEmail(
+      applicationData.email,
+      this.applicantSubject,
+      renderedApplicantBodies
+    ).catch(err => this.rollbar.error(`Error sending applicant email: ${err}`));
```
fionawhim: I’d recommend passing `err` directly to `this.rollbar.error` so that the stack trace and stuff all get sent.

(Here and below)

services-js/commissions-app/src/server/services/Email.ts
```diff
@@ -0,0 +1,109 @@
...
+   * Use Postmark to send an email. The user does not need to be notified
+   * if there is an error.
+   */
+  private sendConfirmationEmail(
+    emailTo: string,
+    subject: string,
+    renderedBodies
+  ): Promise<void> {
+    return new Promise((resolve, reject) => {
+      PostmarkClient.sendEmail(
```
fionawhim: Should this be `this.postmarkClient`?

services-js/commissions-app/src/server/services/Email.ts
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Convert array of applied-for commissions into a human-readable list of
+ * names, and sort it alphabetically.
+ *
+ * @param {Commission[]} commissionsList
+ * @param {string[]} commissionIds
+ */
+export function namesFromIds(commissionsList, commissionIds) {
+  return commissionIds
+    .reduce((acc, id) => {
```
fionawhim: This could also just be a `map` since you don’t need to do anything interesting with the accumulator.

yarn.lock
```diff
@@ -13433,6 +13504,20 @@ mjml-core@^4.1.1:
     mjml-parser-xml "^4.1.1"
     mjml-validator "^4.1.1"
 
+mjml-core@^4.2.0:
```
fionawhim, line 13507: When we get into birth certs we should upgrade registry-certs to this version of mjml as well.

fionawhim (changes requested):

services-js/commissions-app/src/server/email/EmailContent.ts
jessicamarcus: Thanks for the catch; not sure why I wasn’t...

services-js/commissions-app/src/server/services/Email.ts
jessicamarcus: Thanks for catching!

fionawhim (approved):

fionawhim:
Approved after the stuff we talked about in-person!

PR #116, opened 2018-10-22
Adds Google Analytics to commissions-app

fionawhim:
Makes a SiteAnalytics class and Gtag and GA subclasses in
next-client-common. Links it into the existing RouterListener to send
page updates as the user clicks around.

Does not yet update registry-certs to use these new classes.


jessicamarcus (approved):
lgtm!

PR #118, opened 2018-10-25
Factors out registry-certs accessibility

fionawhim:
- Converts registry-certs to use next-client-common’s RouterListener
   and SiteAnalytics implementations.
 - Adds registry-certs’s e-commerce methods to GaSiteAnalytics.
 - Adds timing reporting to RouterListener and the SiteAnalytics
   implementations.
 - Adds registry-certs’s Accessibility class to next-client-common as
   ScreenReaderSupport. Hooks it in to the common RouterListener.
 - Adds ScreenReaderSupport to commissions-app and access-boston as
   well, since it’s built-in now.

Includes additional accessibility tweaks:
 - Changes StatusModal to role="alert" since it doesn’t have dialog form
   controls, and adds aria-live so some screen readers will announce it.
 - Changes Checkbox to take all input attributes. Moves the label
   content into children so that it doesn’t conflict with the title
   attribute.
 - Makes CollapsibleSection’s state apparent in a screen reader by
   changing the button label. Adds the title as the aria-label to the
   section so that it’s still announced.
 - Updates SectionHeader to expose an ID so that it can be used in
   aria-labelledby attributes.
 - Makes all <hr> squiggles be aria-hidden.
 - Changes a lot of things to regions.
 - Changes commissions checkboxes to sound simpler with aria-labels,
   removing some "group"s.
 - Announces submission status for apply form


modules-js/react-fleet/src/sectioning-elements/CollapsibleSection.tsx
```diff
@@ -155,12 +158,18 @@ export default class CollapsibleSection extends React.Component<Props, State> {
...
       className: `sh-title ${HEADER_STYLING}`,
     };
 
+    // We use aria-label for the region so that we can give the <button> a label
+    // that explains whether it’s expanding or collapsing.
     return (
-      <section className={this.props.className}>
+      <section
+        className={this.props.className}
+        role="region"
```
jessicamarcus: `<section>` is automatically treated as a region as long as it has an `aria-label / aria-labelledby` attribute. ref: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Region_role#Accessibility_concerns
```suggestion

```

modules-js/react-fleet/src/sectioning-elements/CollapsibleSection.tsx
```diff
@@ -116,15 +115,19 @@ export default class CollapsibleSection extends React.Component<Props, State> {
...
-      return this.props.title;
+      return <span aria-hidden>{this.props.title}</span>;
     } else {
       return (
         <button
           type="button"
           className={BUTTON_STYLING}
           aria-expanded={this.state.expanded}
+          aria-controls={`cs-${this.props.title.replace(/\s+/g, '-')}`}
+          aria-label={this.state.expanded ? 'Hide section' : 'Show section'}
```
jessicamarcus: The `aria-expanded` attribute already describes whether the section is collapsed or expanded, so this isn’t necessary. This is a nice writeup I’ve used as guidance more than once: https://inclusive-components.design/collapsible-sections/
```suggestion
          
```

modules-js/react-fleet/src/components/StatusModal.tsx
```diff
@@ -47,12 +47,13 @@ export default function StatusModal({
...
-      role="alertdialog"
-      aria-labelledby="statusModal"
-    >
-      <div className={`md-c br br-t400 ${MODAL_STYLE}`} role="document">
+    <div className={`md ${waiting ? WAITING_MODAL_STYLE : ''}`}>
+      <div
+        className={`md-c br br-t400 ${MODAL_STYLE}`}
+        role="alert"
+        aria-labelledby="statusModal"
+        aria-live="assertive"
```
jessicamarcus: An element with `role="alert"` is treated/behaves in the same way as `aria-live="assertive"`. Having both can be problematic: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions#Preferring_Specialized_Live_Region_Roles
```suggestion
        
```

services-js/access-boston/src/client/device-verification/DeviceVerificationForm.tsx
```diff
@@ -75,7 +75,7 @@ export default function DeviceVerificationForm(props: Props) {
             cell phone you can use a personal email address instead.
           </p>
 
-          <hr className="hr hr--sq" />
+          <hr className="hr hr--sq" aria-hidden />
```
jessicamarcus, line 78: I’m not sure this needs to be hidden, because it _does_ play a semantic role to the document structure? I don’t know that it matters either way, tho.

services-js/commissions-app/src/client/CommissionsListSection.tsx
```diff
@@ -187,21 +196,21 @@ function renderCommission(
     <li style={{ listStyleType: 'none' }} key={uid} className="m-b500">
       <Checkbox
         name="commissionIds"
+        aria-label={commission.name}
```
jessicamarcus, line 199: The `<Checkbox>` component wraps its children in a `<label>`, so you could drop this`aria-label` and remove the `aria-hidden` attribute from the sub-`<span>`, and end up with the same information in the accessibility tree (and a bit less markup).
```suggestion
        
```

modules-js/next-client-common/src/ScreenReaderSupport.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
```
jessicamarcus, line 1: This is a longer conversation for another time - and I need to think more on this, and how our apps’ UIs have been assembled - but would this component be providing anything different or additional from what we’d provide a sighted user in the form of an alert or notification of some kind? I’m inclined to prefer a single element that provides the same information to any user, with the ability to add addt’l affordances for screen readers if appropriate.

jessicamarcus (changes requested):
Let’s sit down and talk+brainstorm together tomorrow! Some things in the realm of a11y are totally straightforward (i.e. color contrast, text size), but the bulk of it can get really philosophical and I’ve found that the best solutions arise from the varied perspectives and experiences collaboration with others provides... plus, it’s a particularly attractive topic of nerdery for me, lol.

modules-js/react-fleet/src/sectioning-elements/CollapsibleSection.tsx
fionawhim: Will do.

modules-js/react-fleet/src/sectioning-elements/CollapsibleSection.tsx
fionawhim: Reverting my changes here. I had VoiceOver set to least-verbose and it had stopped reading "expanded" / "collapsed" automatically so I thought those weren’t working.

modules-js/react-fleet/src/components/StatusModal.tsx
fionawhim: Updated to having a static `role="alert"` container.

jessicamarcus (approved):

modules-js/next-client-common/src/ScreenReaderSupport.tsx
fionawhim: Is this sorted out to your satisfaction?

PR #123, opened 2018-10-26
Adds a form-common module for HTML form bits

fionawhim:
- Adds a makeFormData helper so we don't need to use the FormData(form)
   constructor, which has bugs w/ Safari and empty file inputs.
 - Moves in the PHONE_REGEXP since it's a good common place for it.


jessicamarcus (approved):
This will be super convenient for future-us! 😃

PR #128, opened 2018-10-29
Changes "prepare" to only build modules

fionawhim:
No longer builds all of the services when doing a `yarn install`, since
if you’re working on a particular app you’ll be building it anyway, and
apps don’t need to be built to have their tests run.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,8 +21,8 @@
   },
   "scripts": {
     "lint": "eslint .",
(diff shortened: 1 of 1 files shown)


fionawhim:
I think this is a good idea.

jessicamarcus (approved):
It _is_ a good idea. 🎉

PR #129, opened 2018-10-29
Runs prebuild before service tests

fionawhim:
prebuild creates some pregenerated files that are needed by tests.

Fixes the broken develop branch build, since without the pre-building
off apps in "prepare" it didn't have these generated files.


jessicamarcus (approved):
lgtm!

PR #138, opened 2018-11-07
README.md and .env.sample updates for first timers

fionawhim:
Makes it clearer all the Slack app settings you need to put in.

Also enables logging for all environments.


jessicamarcus (approved):
Looks great, thanks!

PR #144, opened 2018-11-09
Removes template test parallelization

fionawhim:
The template tests do Yarn installs, which can interfere with other
tests if they're running at the same time.

This runs the template tests separately on CI, and not at all during
pushes. Changes to use --concurrency lerna flag to make only one run at
a time, so we can remove the lockfile hacks that previously prevented
the templates from stomping each other.


jessicamarcus (approved):
💯

PR #145, opened 2018-11-09
Removes Travis from the staging process

fionawhim:
Listens to GitHub webhooks directly so that the staging deployment can
happen immediately. (We don't run tests for staging deploys, so Travis
is unnecessary.)


jessicamarcus (approved):
🎉

PR #154, opened 2018-11-14
Upgrades 311 to Next 7 / Babel 7

fionawhim:
Fix for the build failures that happen when you have various versions of
Babel and Jest transformations running around.


jessicamarcus (approved):
💯

PR #155, opened 2018-11-14
Updates Dockerfile for Monorepo deploy

fionawhim:
Deletes unused old deployment files.

Also fixes use of this.props in _document for Next 7.


jessicamarcus (approved):
👍

PR #157, opened 2018-11-14
Adds tolerance for no services changing

fionawhim:
- report-updated-services won't ping if there are no changed services
 - DeploymentInteraction ignores notifications with no services

Also adds "fullName" and cleans up some more things around services +
variants to fix builds being reported as "OUTDATED".


jessicamarcus (approved):
lgtm!

PR #162, opened 2018-11-16
Ping templates for Access Boston

jessicamarcus:
Templates from HM that will be hosted by the PingFederate server, rather than our React app. I tried to stick to markup and css as much as possible, leave sections of what I _believe_ to be unused boilerplate in place,  and didn’t really touch the js at all.

**Template change notes:**
- Changed a11y-unfriendly `meta name="viewport"` tag
- Brought in patterns.boston.gov stylesheet
- Removed and replaced custom CSS
- Reworked markup to use fewer and more semantic elements when possible
- Converted `<a>` “buttons” into button elements
- Added `autocomplete` attributes to text fields for the benefit of password managers


services-js/access-boston/src/server/access-boston.ts
```diff
@@ -276,6 +281,76 @@ async function addGraphQl(
   });
 }
 
+async function addVelocityTemplates(server: HapiServer) {
+  server.route({
```
fionawhim, line 285: This is a good candidate for generalization. You could make an array or object with the path / templatePath values in it, and iterate, calling `route` for each of them, since all 4 are exactly the same.

Just a suggestion, though, since this is dev code and therefore it's not tremendously important to remove the duplication.

services-js/access-boston/src/server/ping-templates/mockData.ts
```diff
@@ -0,0 +1,76 @@
...
+    size: () => 0,
+  },
+  loginFailed: false,
+  adapterIdField: 'adapterId Field',
+  adapterId: 'adapterId',
+  localizedMessageResolver: {
+    strings: {
+      locale: 'en',
+    },
+    // @ts-ignore
```
fionawhim: why do you need these? I'd rather you drop in `any` types or something rather than turn off type checking altogether.

fionawhim (approved):

services-js/access-boston/src/server/access-boston.ts
jessicamarcus: I don‘t love the duplication, but wanted to stand these up asap so the template work could be completed and delivered. I may go back later today and refactor after they get emailed out, if you think it’s a good idea/time use.

services-js/access-boston/src/server/ping-templates/mockData.ts
jessicamarcus: The functions being mocked expect two parameters, and I wanted to keep the mocks as similar as possible. Since we’re not using the first parameter in these cases, TS throws a `[parameter] is declared but its value is never read` error. I wasn‘t sure what would be a cleaner solution (but am open to suggestions!).

services-js/access-boston/src/server/access-boston.ts
fionawhim: Totally fine. It's a thing to clean up later if you're back in there or just want the exercise.

services-js/access-boston/src/server/ping-templates/mockData.ts
fionawhim: If you prefix the parameter name with `_` then TS won't give usage complaints.

jessicamarcus: Oh wow, TIL. Thanks!

PR #163, opened 2018-11-16
Moves components from simpler pages to TypeScript

fionawhim:
Also fixes Storybook, and adds Percy processing of Storybook. Moves the
Storybook directory to .storybook, its more standard location, since
that's where our percy-common tool expects it.


fionawhim:
I'm looking into the Percy regressions.

jessicamarcus:
I accidentally approved *QuestionsPane: No Metadata* and *QuestionsPane: No Description*, though the placeholder image for "upload photo" is not loading... I'm hanging back from Percy for now.

fionawhim:
OK, the SVGs are back. I'm not 100% sure how Percy picked the images to compare against, but they're quite old. (This is probably due to this is the first Percy run in the monorepo.)

jessicamarcus:
Is the position of the yellow bottom bar correct here?

https://percy.io/CityOfBoston/311/builds/1179410/view/74667939/1280?mode=diff&browser=firefox&snapshot=74667939

fionawhim:
Yeah, I think so. It’s `position: fixed` so that always comes across a bit off-looking in Percy.

jessicamarcus (approved):
lgtm!

PR #166, opened 2018-11-20
Removes 311’s Store and reworks dependencies

fionawhim:
Switches to the PageDependencies pattern from registry-certs and
commissions-app. Mostly a mechanical transformation that breaks
out the Store’s fields into explicit props.

Puts values that were previously on the request ("RequestAdditions") in
runtime configs or, for languages, the App component.

Anything that touches LocationMap has a lot of dependencies because that
class is used both for the search UI and the location-picking.
Additionally, for mobile devices, LocationMap is served out of the
LocationPopUp, so all of its dependencies need to be threaded to there.


fionawhim:
This is likely the last giant PR for the moment.

jessicamarcus (approved):
lgtm! 🎉

PR #170, opened 2018-11-26
Changes staging deploys to update task definitions

fionawhim:
Allows us to roll back staging deploys when they fail. Previously we
just updated the container tag and forced a deploy, but when things
didn’t start up the service would flap indefinitely, eventually eating
up the host’s space.

This makes staging deploys work like production. The only change will be
more task definition versions, which shouldn’t be a big deal.


jessicamarcus (approved):
Sounds good to me.

PR #172, opened 2018-11-27
Converts 311-indexer from Flow to TypeScript

fionawhim:


fionawhim:
This is the last bit of Flow currently in the monorepo.

And yes, the comments for the RxJS stuff should just be "im sorry im sorry im sorry im sorry"

Once I get this in I'm going to switch from RxJS 5 to 6 which cleans up the API a fair amount and should make it slightly more understandable.

jessicamarcus (approved):
Hah - it’s been a while since I’ve looked at any code, but I’d rather see RxJS than JQuery in a codebase any day. 😉

PR #173, opened 2018-11-28
initial commit

jessicamarcus:
Initial pass at adding in the birth certs question workflow. Also:
- updates `.nvmrc`
- refactors PageLayout component for breadcrumb-related reasons
- adds `onKeyDown` prop to TextInput in react-fleet


services-js/registry-certs/client/PageLayout.tsx
```diff
@@ -8,41 +8,42 @@ import DeathCertificateCart from './store/DeathCertificateCart';
 
 import Nav from './common/Nav';
 
+type breadcrumbNavLink = {
```
fionawhim: Types should typically start with a capital letter, here and below.

services-js/registry-certs/client/birth/AnswersSummary.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+// import moment from 'moment';
```
fionawhim: remove commented out?

Also, it's best to avoid moment in client code because it's pretty big. date-fns is a good alternative, or figuring out a strategy for formatting dates on the server.

services-js/registry-certs/client/birth/QuestionsFlow.tsx
```diff
@@ -0,0 +1,344 @@
...
+export type question =
+  | 'forSelf'
+  | 'howRelated'
+  | 'bornInBoston'
+  | 'parentsLivedInBoston'
+  | 'nameOnRecord'
+  | 'dateOfBirth'
+  | 'parentsMarried'
+  | 'parentsNames'
+  | 'endFlow';
```
fionawhim, line 25: good

services-js/registry-certs/client/birth/QuestionsFlow.tsx
```diff
@@ -0,0 +1,344 @@
...
+ *
+ * User will progress to checkout upon completion of this workflow.
+ */
+
+// todo: sort things out to remove all @ts-ignores
+export default class QuestionsFlow extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+
+    // @ts-ignore
```
fionawhim: Please don't use `@ts-ignore`.

services-js/registry-certs/client/birth/QuestionsFlow.tsx
```diff
@@ -0,0 +1,344 @@
...
+      text: 'Registry: Birth, death, and marriage',
+    },
+  ],
+  currentPage: {
+    url: '/birth',
+    text: 'Birth certificates',
+  },
+};
+
+const initialState = {
```
fionawhim: Should be able to type this as `State`

services-js/registry-certs/client/birth/QuestionsFlow.tsx
```diff
@@ -0,0 +1,344 @@
...
+export default class QuestionsFlow extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+
+    // @ts-ignore
+    this.state = initialState;
+  }
+
+  // this.state.answeredQuestions is an ordered record of the questions asked.
+  addToAnsweredQuestions(newQuestion: question): void {
```
fionawhim: I've taken to adding `private` when possible to methods so that we get unused errors if we stop calling them.

services-js/registry-certs/client/birth/QuestionsFlow.tsx
```diff
@@ -0,0 +1,344 @@
...
+    const { answeredQuestions } = this.state;
+
+    answeredQuestions.push(newQuestion);
```
fionawhim: Don’t use `push` on things you get out of state because it mutates them. Better to do something along the lines of:

```
this.setState({
  answeredQuestions: [...answeredQuestions, newQuestion]
});
```

services-js/registry-certs/client/birth/QuestionsFlow.tsx
```diff
@@ -0,0 +1,344 @@
...
+  ): void => {
+    this.setState({ activeQuestion: nextQuestion });
+
+    this.addToAnsweredQuestions(currentQuestion);
+  };
+
+  // When the user selects “back”, we treat that previous question
+  // as unanswered, and return to its screen.
+  stepBackOneQuestion = () => {
+    interface FieldsToClear {
```
fionawhim: it feels odd to have an interface declared inside a method.

Looking below, I think what you want to do is type `fieldsToClear` as `Partial<State>`, which turns all of `State`’s fields optional.

services-js/registry-certs/client/birth/QuestionsFlow.tsx
```diff
@@ -0,0 +1,344 @@
...
+      firstName?: string;
+      lastName?: string;
+      parent1FirstName?: string;
+      parent1LastName?: string;
+      parent2FirstName?: string;
+      parent2LastName?: string;
+    }
+
+    const answeredQuestions: question[] = this.state.answeredQuestions;
+    // @ts-ignore
```
fionawhim: Rather than do this, either be defensive of `pop` returning `undefined` or add `!` to assert that it’s defined.

That being said, you could probably get by with just removing some of the types from the local variables. Since `this.state.answeredQuestions` is a `question[]`, TypeScript will propagate that type correctly.

Typically the only times you want a type on a local variable are:
 * The function you’re calling is returning `any` and you want to coerce it for downstream consumption
 * You're assigning a literal (especially `[]`) and want to make sure it’s type-checked against a known type.

services-js/registry-certs/client/birth/QuestionsFlow.tsx
```diff
@@ -0,0 +1,344 @@
...
+    const prevQuestion: question = answeredQuestions.pop();
```
fionawhim: warning as well that `pop` mutates. You should either use `slice` or operate on a copy of `this.state.answeredQuestions`.

services-js/registry-certs/client/birth/QuestionsFlow.tsx
```diff
@@ -0,0 +1,344 @@
...
+    this.setState({
+      answeredQuestions,
+      activeQuestion: prevQuestion,
+      ...fieldsToClear,
+    });
+  };
+
+  handleTextInput = (event: React.ChangeEvent<HTMLInputElement>): void => {
+    // @ts-ignore
+    this.setState({
```
fionawhim, line 160: Rather than `@ts-ignore` in this case, cast the argument to `setState` to `any`. That way it’s clearer exactly what is being hand-waved.

services-js/registry-certs/client/birth/QuestionsFlow.tsx
```diff
@@ -0,0 +1,344 @@
...
+      [event.target.name]: event.target.value,
+    });
+  };
+
+  handleForSelf = (event: React.ChangeEvent<HTMLInputElement>): void => {
+    const value = JSON.parse(event.target.value);
+
+    this.setState({ forSelf: value }, () =>
+      this.proceedToNextQuestion(
```
fionawhim: I don't think you need to do this in a callback unless `proceedToNextQuestion` explicitly needs to read out the value of `forSelf`. (here and below)

services-js/registry-certs/client/birth/QuestionsFlow.tsx
```diff
@@ -0,0 +1,344 @@
...
+        'bornInBoston',
+        value === 'yes' ? 'nameOnRecord' : 'parentsLivedInBoston'
+      )
+    );
+  };
+
+  // This question is only presented if the user selects NO or UNKNOWN for bornInBoston.
+  // If NO for both “born in Boston” and “parents lived in Boston”,
+  // exit the workflow and inform the user.
+  // If UNKNOWN for either question, proceed to next question.
```
fionawhim, line 206: good comments

services-js/registry-certs/client/birth/QuestionsFlow.tsx
```diff
@@ -0,0 +1,344 @@
...
+  | 'grandparent'
+  | 'sibling'
+  | 'spouse'
+  | 'friend'
+  | 'attorney'
+  | 'guardian'
+  | 'other'
+  | string;
+
+export type yesNoUnknownAnswer = 'yes' | 'no' | 'unknown' | string;
```
fionawhim: Do you actually want `string` here? Might be better to coerce the type of `input.value`. Leaving `string` as a possible wildcard may keep us from catching bugs.

services-js/registry-certs/client/birth/QuestionsFlow.tsx
```diff
@@ -0,0 +1,344 @@
...
+      this.proceedToNextQuestion(
+        'parentsLivedInBoston',
+        value === 'no' && this.state.bornInBoston === 'no'
+          ? 'endFlow'
+          : 'nameOnRecord'
+      )
+    );
+  };
+
+  handleNameOnRecord = (): void => {
```
fionawhim: You don’t need all these `: void`s.

services-js/registry-certs/client/birth/questions/BornInBoston.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+
+import FieldsetComponent from './FieldsetComponent';
+import YesNoUnsureComponent from './YesNoUnsureComponent';
+
+interface Props {
+  forSelf: boolean;
+  handleChange: (event) => void;
```
fionawhim: Should put a type on the `event` argument, here and below.

services-js/registry-certs/client/birth/questions/DateOfBirth.tsx
```diff
@@ -0,0 +1,61 @@
...
+  handleTextInput: (event) => void;
+  handleProceed: () => void;
+  handleStepBack: () => void;
+}
+
+export default function DateOfBirth(props: Props): JSX.Element {
+  const handleEnterKeypress = (event): void => {
+    const key = event.key || event.keyCode;
+
+    if (key === 'Enter' || key === 13) props.handleProceed();
```
fionawhim: In general use `{}`s for all `if`s.

I think React normalizes the `key` property of events.

services-js/registry-certs/client/birth/questions/EndFlow.tsx
```diff
@@ -0,0 +1,76 @@
...
+              duties, in the opinion of the city clerk or the commissioner of
+              public health, as the case may be, entitle them to the information
+              contained therein.
+            </p>
+          )}
+        </>
+      );
+    }
+
+    return (
```
fionawhim, line 40: It feels more balanced to put this in an `else`, since the `canProceed` check isn’t really a short-circuit.

services-js/registry-certs/client/birth/questions/FieldsetComponent.tsx
```diff
@@ -0,0 +1,83 @@
...
+  handleStepBack?: () => void;
+  children: React.ReactNode;
+}
+
+/**
+ * Container component to provide layout for a single question screen,
+ * as well as “back” and “next question” buttons if their related handlers
+ * are passed in as props to this component.
+ */
+export default function FieldsetComponent(props: Props): JSX.Element {
```
fionawhim, line 23: Good!

services-js/registry-certs/client/birth/questions/FieldsetComponent.tsx
```diff
@@ -0,0 +1,83 @@
...
+            type="button"
+            onClick={props.handleStepBack}
+            className={`btn ${BACK_BUTTON_STYLING}`}
+          >
+            Back
+          </button>
+        )}
+
+        {props.handleProceed && (
+          <button type="button" onClick={props.handleProceed} className="btn">
```
fionawhim, line 42: TODO: disable this as necessary

services-js/registry-certs/client/birth/QuestionsFlow.tsx
```diff
@@ -0,0 +1,344 @@
...
+
+  handleTextInput = (event: React.ChangeEvent<HTMLInputElement>): void => {
+    // @ts-ignore
+    this.setState({
+      [event.target.name]: event.target.value,
+    });
+  };
+
+  handleForSelf = (event: React.ChangeEvent<HTMLInputElement>): void => {
+    const value = JSON.parse(event.target.value);
```
fionawhim: Either add `boolean` type to `value` explicitly, or change to `event.target.value === 'true'`

fionawhim (changes requested):
Overall I like the structure you have set up for the questions.

There are a handful of things to fix here; be on the lookout for some things that I mentioned once but should be addressed everywhere they show up.

services-js/registry-certs/client/birth/AnswersSummary.tsx
jessicamarcus: No problem - I was going to use Moment because it was already being used in death certs, but will skip it and use Intl.DateTimeFormat instead. 😃

services-js/registry-certs/client/birth/QuestionsFlow.tsx
jessicamarcus: I know - there’s a todo at the top of the class for me to go back and figure out why tsc was unhappy.

services-js/registry-certs/client/birth/questions/DateOfBirth.tsx
jessicamarcus: ohhhh of course: React uses its SyntheticEvents. Thanks!

services-js/registry-certs/client/birth/questions/EndFlow.tsx
jessicamarcus: Thanks - it’s more explicit of intention as well. 👍

services-js/registry-certs/client/birth/QuestionsFlow.tsx
fionawhim: Still need to address this mutation.

services-js/registry-certs/client/birth/questions/BornInBoston.tsx
fionawhim: Ping on these.

fionawhim (approved):

PR #174, opened 2018-11-29
Make indexer sensitive to meta failures

fionawhim:
This throws an error when a meta message is a failure, such as when
Salesforce errors during a long-poll reconnect. Previously these were
ignored and we just stopped receiving messages.

Updates the error handling to complete the RxJS stream rather than
immediately quit, which allows existing processing to cleanly complete.

Also updates to latest bugfix version of Comet 3.1 line and adds http
proxy support in more places for better debugging with Charles.

### DIFF
--- a/services-js/311-indexer/server/vendor/cometd-nodejs-client.js
+++ b/services-js/311-indexer/server/vendor/cometd-nodejs-client.js
@@ -1,0 +1,186 @@
+// vendored 11/28/19 by Fin Hopkins from https://github.com/cometd/cometd-nodejs-client
+// under the terms of the Apache license.
+//
+// Modified to add the "makeAgent" function that allows for an HTTP proxy.
+
+module.exports = {
+  adapt: function(makeAgent) {
+    var url = require('url');
+    var httpc = require('http');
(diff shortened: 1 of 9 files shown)


services-js/311-indexer/server/vendor/cometd-nodejs-client.js
```diff
@@ -0,0 +1,186 @@
+// vendored 11/28/19 by Fin Hopkins from https://github.com/cometd/cometd-nodejs-client
```
jessicamarcus:
```suggestion
// vendored 11/28/18 by Fin Hopkins from https://github.com/cometd/cometd-nodejs-client
```

jessicamarcus: don’t scare me like that! 😉

jessicamarcus (approved):

services-js/311-indexer/server/vendor/cometd-nodejs-client.js
fionawhim: lol, good catch. :)

PR #178, opened 2018-11-29
Keeps indexer from skipping same-minute cases

fionawhim:
The imported 311 cases only have 1-minute granularity, so our old
strategy of always subtracting 1 second between API loads was in danger
of skipping at a higher chance that we were expecting.

This updates things to make sure to re-load the last minute to get any
other cases that happened that minute, and adds the safety check to make
sure we don't keep requesting the same minute over and over again.

Also takes the opportunity to remove the "selectionFilter" argument to
mergeMap, which is deprecated in RxJS 6, replacing it with a map call to
transform the output of the indexing mergeMap.


jessicamarcus (approved):
Looks great to me!

PR #179, opened 2018-11-29
Mechanical transform from RxJS 5 to 6

fionawhim:
Removes a bunch of Observable methods in favor of the "pipe" function
and operator functions.

Also refactors where we set the parallelism for the classifier.


jessicamarcus (approved):
👍

PR #180, opened 2018-11-30
Lowers 311 indexer loading date range

fionawhim:
Around DST our 90 day calculation was different from Salesforce’s and it
errored out. 88 days gives us a bit of slop so that we’re guaranteed
under 90 for the range.

Also fixes some rxjs imports, which need the "* as" construction.


jessicamarcus (approved):
👍

PR #184, opened 2018-12-03
Google Analytics for B&C Search

mmcgowanBOS:
- Added Google Analytics node module.
- Added front end js to REACT app.

### DIFF
--- a/services-js/commissions-search/package.json
+++ b/services-js/commissions-search/package.json
@@ -12,7 +12,6 @@
     "react": "16.6.0",
     "react-apollo": "^2.1.9",
     "react-dom": "16.6.0",
-    "react-ga": "^2.5.6",
     "react-scripts": "1.1.4"
   },
   "scripts": {
--- a/services-js/commissions-search/package.json
+++ b/services-js/commissions-search/package.json
@@ -12,6 +12,7 @@
     "react": "16.6.0",
     "react-apollo": "^2.1.9",
     "react-dom": "16.6.0",
+    "react-ga": "^2.5.6",
     "react-scripts": "1.1.4"
   },
   "scripts": {
--- a/services-js/commissions-search/src/App.js
+++ b/services-js/commissions-search/src/App.js
@@ -1,10 +1,12 @@
 /* eslint-disable no-unused-vars */
 import React from 'react';
 import update from 'immutability-helper';
-
 import Search from './Search';
 import FacetList from './FacetList';
 import ResultList from './ResultList';
+import ReactGA from 'react-ga';
+
+ReactGA.initialize('UA-2187282-27');
 
 class App extends React.Component {
   /*
@@ -50,6 +52,30 @@ class App extends React.Component {
     this.setState({ currentSeats: event.target.value });
   };
 
+  findCheckedBoxes4GA = event => {
+    var chk_arr = document.getElementsByClassName('cb');
+    var chklength = chk_arr.length;
+    for (var k = 0; k < chklength; k++) {
+      var checkit = chk_arr[k].firstChild.checked ? true : false;
+      if (checkit === true) {
+        var children = chk_arr[k].childNodes;
+        this.sendEvent2GA(
+          'Search Filter',
+          'Commissions Search',
+          children[1].innerText
+        );
+      }
+    }
+  };
+
+  sendEvent2GA = (actionV, categoryV, labelV) => {
+    ReactGA.event({
+      category: categoryV,
+      action: actionV,
+      label: labelV,
+    });
+  };
+
   handleFacetSubmit = event => {
     event.preventDefault();
 
@@ -57,6 +83,9 @@ class App extends React.Component {
       submittedAreas: this.state.currentAreas,
       submittedSeats: this.state.currentSeats,
     });
+
+    //send search terms to GA Event Tracking
+    this.findCheckedBoxes4GA();
   };
 
   handleKeywordChange = event => {
@@ -69,6 +98,12 @@ class App extends React.Component {
     this.setState({
       submittedKeywords: this.state.currentKeywords,
     });
+    //send search terms to GA Event Tracking
+    this.sendEvent2GA(
+      'Search Box',
+      'Commissions Search',
+      this.state.currentKeywords
+    );
   };
 
   render() {
--- a/services-js/commissions-search/src/App.js
+++ b/services-js/commissions-search/src/App.js
@@ -4,9 +4,6 @@ import update from 'immutability-helper';
 import Search from './Search';
 import FacetList from './FacetList';
 import ResultList from './ResultList';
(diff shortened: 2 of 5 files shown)


services-js/commissions-search/src/App.js
```diff
@@ -1,10 +1,12 @@
 /* eslint-disable no-unused-vars */
 import React from 'react';
 import update from 'immutability-helper';
-
 import Search from './Search';
 import FacetList from './FacetList';
 import ResultList from './ResultList';
+import ReactGA from 'react-ga';
+
+ReactGA.initialize('UA-2187282-27');
```
fionawhim: It’s not good to hard-code values like this into source code.

In theory configuration values like this should be made available by something in our Drupal template. See how we're doing that for the API key here: https://github.com/CityOfBoston/digital/blob/develop/services-js/commissions-search/src/index.js#L25

Though since you seem to want to use the same GA property as Boston.gov already has, do you really need `ReactGA`, as opposed to just assuming that the Google Analytics stuff is already loaded?

I'm worried that this will lead to double-counting.

services-js/commissions-search/src/App.js
```diff
@@ -50,13 +52,40 @@ class App extends React.Component {
     this.setState({ currentSeats: event.target.value });
   };
 
+  findCheckedBoxes4GA = event => {
+    var chk_arr = document.getElementsByClassName('cb');
```
fionawhim: A few style things:

 - Use camelCase for JavaScript identifiers.
 - Don’t use `var`. Use `const` by default, `let` if you need to.

services-js/commissions-search/src/App.js
```diff
@@ -50,13 +52,40 @@ class App extends React.Component {
...
+    var chklength = chk_arr.length;
+    for (var k = 0; k < chklength; k++) {
+      var checkit = chk_arr[k].firstChild.checked ? true : false;
```
fionawhim: This data should already be present in the component’s state, so you should get it from there rather than inspecting HTML elements.

Also, avoid `document`-based selection in general in React, since you could be grabbing input elements from other places on the page. If you reeeeeaaaaalllly need native DOM operations, use a Ref to get access to an HTML element specifically owned by your component.

services-js/commissions-search/src/App.js
```diff
@@ -50,13 +52,40 @@ class App extends React.Component {
...
+        this.sendEvent2GA(
+          'Search Filter',
+          'Commissions Search',
+          children[1].innerText
+        );
+      }
+    }
+  };
+
+  sendEvent2GA = (actionV, categoryV, labelV) => {
```
fionawhim, line 72: good encapsulation.

fionawhim (changes requested):

services-js/commissions-search/src/App.js
```diff
@@ -53,12 +50,12 @@ class App extends React.Component {
   };
 
   findCheckedBoxes4GA = event => {
-    var chk_arr = document.getElementsByClassName('cb');
-    var chklength = chk_arr.length;
-    for (var k = 0; k < chklength; k++) {
-      var checkit = chk_arr[k].firstChild.checked ? true : false;
-      if (checkit === true) {
-        var children = chk_arr[k].childNodes;
+    const chkArr = document.getElementsByClassName('cb');
```
fionawhim: I know we talked about commenting this code (to explain why you’re not using state) and using something lest fragile than `firstChild` and `children[1]`. 

Is there a reason you didn't go that route?

mmcgowanBOS: I was waiting for Travis to finish and then going to make comments in GitHub and address the questions / comments there.

services-js/commissions-search/src/App.js
mmcgowanBOS: Removed GA module. Now using dataLayer Object to push data to GA. 
*Note, this approach does not work with existing Google Tag Manager (GTM) included on the Drupal (or any other) template page UNLESS specific Tags, Triggers, and Variables are set up to integrate properly in GTM.

services-js/commissions-search/src/App.js
mmcgowanBOS: sure.

services-js/commissions-search/src/App.js
mmcgowanBOS: Per our meeting / conversation, this was / is not available in the component state and would have to be altered using a rather buried Apollo set up.

fionawhim: Can you put that in the code as a comment?

It's much more likely to be found there when someone needs to look back on the code than digging up the old PR.

paylakatel (approved):

services-js/commissions-search/package.json
```diff
@@ -12,6 +12,7 @@
     "react": "16.6.0",
     "react-apollo": "^2.1.9",
     "react-dom": "16.6.0",
+    "react-ga": "^2.5.6",
```
fionawhim: Remove this, since it's not being used, and be sure to re-run `yarn install` so it gets taken out of `yarn.lock`

services-js/commissions-search/src/App.js
```diff
@@ -30,33 +31,63 @@ class App extends React.Component {
...
+    this.setState({ checkedNames });
   };
 
   handleOptionChange = event => {
     this.setState({ currentSeats: event.target.value });
   };
 
+  //get checkbox names from state and send to GA
+  findCheckedBoxes4GA = event => {
+    const chkArr = { ...this.state.checkedNames };
```
fionawhim, line 65: You don't need to splat this into a new Object.

Also, `Arr` is not a good name for this since it's an object, not an array.

services-js/commissions-search/src/FacetList.js
```diff
@@ -25,6 +25,7 @@ function renderAreaCheckbox(props, policyArea) {
     <Checkbox
       key={`checkbox-area-${policyArea.id}`}
       name={policyArea.id}
+      nameval={labelText}
```
fionawhim, line 28: I'd recommend using `data-` for custom attributes. (E.g. `data-policy-name={labelText}`)

services-js/commissions-search/src/App.js
mmcgowanBOS: I ended up adding an attribute to the checkbox input elements in the FacetList.js file that displays the label value. Having these values available directly in the DOM we now can easily access them and add to REACT state in the App.js file. From here we can iterate over the new state Object (checkedNames) upon submission of the filter and send to GA.

services-js/commissions-search/src/App.js
mmcgowanBOS: See comments above for updates to this.

fionawhim:
Approved per Kayla’s request.

PR #185, opened 2018-12-03
Converts 311 from Opbeat to Rollbar

fionawhim:


fionawhim:
Diffs in the service is mostly removing the transaction code, which was in a `try`/`finally` block, so those look messier than they really are.

jessicamarcus (approved):
Looks great to me!

PR #187, opened 2018-12-04
Fixes for 311 Live Agent functionality

fionawhim:
- Sets the Live Agent button ID so that the LiveAgent service object
   can use it correctly
 - Encapsulates all code needed to interact with Salesforce into the
   LiveAgent service
 - Changes button ID configuration to happen in the constructor, since
   changing the button ID at runtime is not something we need to do

Fixes CityOfBoston/311#934
Fixes CityOfBoston/311#935


jessicamarcus (approved):
lgtm!

PR #190, opened 2018-12-05
Adds simple cache support to fetchGraphql

### LINKED ISSUE #188: Cache service descriptions in the client
Since removing the generic loopbackGraphql cache, the 311 webapp is making more calls to get the details for a service.

We should cache those results on the client to reduce traffic.

fionawhim:
Caching is done by passing in a cache object when creating the
fetchGraphql function instance, and providing a cache key any time you
want to potentially read from / write to the cache.

Makes it straightforward to cache requests made during server-side
rendering so that they’re not re-fetched in the browser.

Fixes #188

### DIFF
--- a/modules-js/next-client-common/src/next-client-common.ts
+++ b/modules-js/next-client-common/src/next-client-common.ts
@@ -44,6 +44,8 @@ export interface RuntimeConfig {
 
 export type QueryVariables = { [key: string]: any };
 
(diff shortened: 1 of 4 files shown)


fionawhim:
I put this on staging for registry-certs and commissions-app and it looks like they're still working.

jessicamarcus (approved):

PR #194, opened 2018-12-05
Updates TestCafe to 0.23.2

fionawhim:
Seems to add better reliability (retries) and faster Chrome exiting on
local OSX.

Also lays the groundwork for 311 testcafe tests.


jessicamarcus (approved):
👍

PR #197, opened 2018-12-07
Changes the browser URL for 311 suggest pane

fionawhim:
Next.js treats any location changes where the browser URL looks the same
as replaceState rather than pushState, so we need to change the URL to
make the back button work.


jessicamarcus (approved):
👍

PR #198, opened 2018-12-07
Fixes 311 Storyshots

fionawhim:
The stubbing of jest functions was keeping it from working. We don’t
seem to need those anymore, so I removed them.


jessicamarcus (approved):
Sounds good to me. 👍

PR #201, opened 2018-12-07
Adds sourcemap generation for registry-certs

fionawhim:
Useful so that Rollbar can show us better stack traces for production
errors. Currently we’re not doing any fancy pre-loading of sourcemaps to
Rollbar. If it turns out we’d like that extra reliability we can
implement it.

And we don’t care that it’s all public because we’re open source
anyway!

Note we have to use the 0.0.4 canary for the source map plugin because
it’s the version that supports Terser, rather than UglifyJS. (Next
switched at some point.)


jessicamarcus (approved):
Yessssssssss! 🎉

PR #202, opened 2018-12-07
Waits for the React app to start in TestCafe

fionawhim:
Part of our TestCafe flakiness is that the forms are getting filled out
before React is done starting up, so when it does the form fields get
wiped out by the empty values.

This makes sure that the isReactRunning function is returning true
before we exit out of the logIn helper.


jessicamarcus (approved):
💯

PR #203, opened 2018-12-07
Adds resiliency when localStorage is not available in the browser

jessicamarcus:
- Prevents uncaught error if localStorage is not available in the browser
- Hides “save contact info on computer” option on shipping information page in checkout

### DIFF
--- a/services-js/registry-certs/client/death/checkout/ShippingContent.tsx
+++ b/services-js/registry-certs/client/death/checkout/ShippingContent.tsx
@@ -101,6 +101,7 @@ export default class ShippingContent extends React.Component<Props, State> {
 
     const {
       shippingIsComplete,
+      localStorageAvailable,
       info: {
         storeContactAndShipping,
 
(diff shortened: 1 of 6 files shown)


services-js/registry-certs/client/death/checkout/ShippingContent.tsx
```diff
@@ -387,20 +388,22 @@ export default class ShippingContent extends React.Component<Props, State> {
...
-                  value="true"
-                  checked={storeContactAndShipping}
-                  {...this.fieldListeners('storeContactAndShipping')}
-                  className="cb-f"
-                />
-                <span className="cb-l">
-                  Save contact and shipping info on this computer
-                </span>
-              </label>
+              {localStorageAvailable && (
```
fionawhim, line 391: Need to do the same thing in `PaymentContent.tsx`

fionawhim (approved):
LGTM after adding the same check to `PaymentContent.tsx`

PR #205, opened 2018-12-10
Fixes for mobile UI for 311 search

fionawhim:
- Adds @observable back in on some HTML elements because @computed
   functions rely on them and need to re-compute when they become
   not-null.
 - Just removes hover handlers when doing the mobile UI so that tapping
   on a search result will go to it, rather than (potentially) show the
   map pop-up and scroll it (which then causes another search, which
   changes the list).

Adds an Elasticsearch fake so that this part of the UI can be exercised
without AWS credentials.

### DIFF
--- a/services-js/311/server/services/__fixtures__/elasticsearch.json
+++ b/services-js/311/server/services/__fixtures__/elasticsearch.json
@@ -1,0 +1,44 @@
+[
+  {
+    "service_request_id": "200012223",
+    "status": "open",
+    "location": null,
+    "address": "50 Milk St.",
+    "description": "There are large rats here. They’re biting.",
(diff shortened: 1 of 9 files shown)


services-js/311/server/services/__fixtures__/elasticsearch.json
```diff
@@ -0,0 +1,44 @@
+[
+  {
+    "service_request_id": "200012223",
+    "status": "open",
+    "location": null,
+    "address": "50 Milk St.",
+    "description": "There are large rats here. They’re biting.",
```
jessicamarcus, line 7: 4c1h

jessicamarcus (approved):
lgtm!

PR #207, opened 2018-12-12
Changes PROD_311_* env vars to OPEN311_

fionawhim:
Saying "prod" is a legacy, confusing thing.


jessicamarcus (approved):
Definitely less confusing!

PR #209, opened 2018-12-13
Updates questions text; adds progress bar; general code cleanup

jessicamarcus:
- Text of questions has been updated to the latest copy
- Added basic progress bar; this can be styled and used as defqault/for small viewports, along with a more complex one based on final UI design
- Moved some shared types and values into their own files


services-js/registry-certs/client/birth/constants.ts
```diff
@@ -0,0 +1,18 @@
+import { BreadcrumbNavigation } from '../PageLayout';
+
+export const BREADCRUMB_NAV_LINKS: BreadcrumbNavigation = {
```
fionawhim, line 3: 👍

fionawhim (approved):
I'm not a super fan of the "Example: Mother" / "Example: Father" bit both from an awkward language perspective as well as a homosexual one, but that's a separate discussion that we don't have to hold up this PR for.

jessicamarcus:
Yeah, I am not a fan of the language either; I’ll send you the link to the working document so you can offer feedback.

PR #213, opened 2018-12-14
Allows partial case numbers for legacy cases

### LINKED ISSUE #888: Allow partial case # searches
According to Alex, call takers will give out just the last 5 digits of a case number. _E.g._ "15503"

If a constituent searches for that number then they should find their case in the search page.

fionawhim:
If the user searches for a 7-digit case ID that’s in the range used for
Lagan, we automatically add the prefix to fill it out to the full Lagan
case ID format.

Fixes CityOfBoston/311#888


jessicamarcus (approved):
Looks much more resilient!

PR #215, opened 2018-12-14
Adds request flow integration test

fionawhim:
- Fills out ArcGISFake to return addresses and units
 - Adds TestCafe tests that start at the home dialog to do search and
   create a new request


jessicamarcus (approved):
lgtm! At first glance, I was scratching my head at how you managed to misspell “modal” so many times... 🤣

PR #217, opened 2018-12-14
Adds mobx state object for birth certs

jessicamarcus:
- Moved question results storage to a mobx object from QuestionsFlow component
- Removed complexity because the question flow is no longer branching
- Added in ReviewRequest page/component
- Added in placeholder for CheckoutFlow page/component
- Reorganized some shared types and styles


services-js/registry-certs/client/birth/QuestionsFlow.tsx
```diff
@@ -10,116 +13,118 @@ import DateOfBirth from './questions/DateOfBirth';
...
-  firstName: '',
-  lastName: '',
-  altSpelling: '',
-  dateOfBirth: '',
-  parentsMarried: '',
-  parent1FirstName: '',
-  parent1LastName: '',
-  parent2FirstName: '',
-  parent2LastName: '',
+  questions: [
```
fionawhim: I don't see the value of this changing, so (if that's true) it would be better as a `const QUESTIONS` or something. Being in state makes it look like it’s supposed to get modified.

services-js/registry-certs/client/birth/QuestionsFlow.tsx
```diff
@@ -10,116 +13,118 @@ import DateOfBirth from './questions/DateOfBirth';
...
 export default class QuestionsFlow extends React.Component<Props, State> {
   state: State = INITIAL_STATE;
 
-  // this.state.answeredQuestions is an ordered record of the questions asked.
-  private addToAnsweredQuestions(newQuestion: Question): void {
-    const answeredQuestions = this.state.answeredQuestions;
-
-    this.setState({ answeredQuestions: [...answeredQuestions, newQuestion] });
-  }
+  private answerQuestion = (question, answers): void => {
```
fionawhim: Can you put type annotations on these arguments?

services-js/registry-certs/client/birth/QuestionsFlow.tsx
```diff
@@ -10,116 +13,118 @@ import DateOfBirth from './questions/DateOfBirth';
...
+  | 'howRelated';
 
-interface Props {}
+interface Props extends Pick<PageDependencies, 'birthCertificateRequest'> {}
 
-interface State extends RequestInformation {
+interface State {
   activeQuestion: Question;
+  questions: Question[];
   answeredQuestions: Question[];
```
fionawhim: Can `answeredQuestions` be removed now?

services-js/registry-certs/client/birth/QuestionsFlow.tsx
```diff
@@ -10,116 +13,118 @@ import DateOfBirth from './questions/DateOfBirth';
...
+        Step {currentIndex + 1}
+      </progress>
+    );
+  }
 
-    this.setState({ ...answers, forSelf });
-    this.proceedToNextQuestion('forSelf', 'bornInBoston');
-  };
+  componentDidMount(): void {
+    // If parent1FirstName has a value, the user has come over from /review
```
fionawhim: I think this logic belongs in the constructor, to set the initial state based on props. Otherwise you get an unnecessary re-render.

services-js/registry-certs/client/birth/ReviewRequest.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+  returnToQuestions = () => {
+    // There is a check in componentDidMount to see if the required field
+    // for the last question has been answered; this will bring the user
+    // back to that last question.
+    Router.push('/birth');
+  };
+
+  goToCheckout = () => {
+    this.props.birthCertificateRequest.setQuantity(
```
fionawhim: Do we want state as an intermediary or should the quantity come from `birthCertificateRequest` directly?

I'm thinking of the case of updating the quantity, then going back to questions, then coming forward to the review and the quantity being different.

services-js/registry-certs/client/birth/QuestionsFlow.tsx
```diff
@@ -10,116 +13,118 @@ import DateOfBirth from './questions/DateOfBirth';
...
 };
 
 /**
  * Guides the user through a number of questions in order to provide the
  * Registry with the information they will need to locate the birth record.
  *
- * User will progress to checkout upon completion of this workflow.
+ * User will progress to /review upon completion of this workflow.
  */
 export default class QuestionsFlow extends React.Component<Props, State> {
```
fionawhim: This should be annotated with `@observer()` if you’re reading MobX values.

services-js/registry-certs/client/birth/QuestionsFlow.tsx
fionawhim: Though I'm also realizing that if you kept `activeQuestion` in `birthCertificateRequest` rather than in this component’s state, you wouldn’t need this awkwardness.

services-js/registry-certs/client/birth/ReviewRequest.tsx
```diff
@@ -0,0 +1,223 @@
...
+}
+
+/**
+ * Component which allows a user to review their request, and update the
+ * quantity of birth certificates they are requesting.
+ *
+ * User can proceed to /checkout, go back to the questions flow, or
+ * clear all information and start over.
+ */
+export default class ReviewRequest extends React.Component<Props, State> {
```
fionawhim: This should also be `@observer()`

services-js/registry-certs/client/birth/ReviewRequest.tsx
```diff
@@ -0,0 +1,223 @@
...
+      lastName,
+      birthDate,
+    } = this.props.birthCertificateRequest.requestInformation;
+
+    return (
+      <>
+        <Head>
+          <title>Boston.gov — Review your record request</title>
+
+          <style>{'.txt-f { border-radius: 0; }'}</style>
```
fionawhim, line 73: Could also do this with Emotion, though I guess I don't have strong feelings.

services-js/registry-certs/client/birth/ReviewRequest.tsx
```diff
@@ -0,0 +1,223 @@
...
+  marginRight: '1rem',
+  fontFamily: 'inherit',
+  fontStyle: 'italic',
+  fontSize: '1rem',
+  background: OPTIMISTIC_BLUE,
+  color: 'white',
+  textAlign: 'right',
+  padding: '0.5rem',
+  '-moz-appearance': 'textfield',
+  '&::-webkit-inner-spin-button': {
```
fionawhim, line 171: Is this the invocation to get rid of the little up/down arrows? TIL.

services-js/registry-certs/client/birth/questions/DateOfBirth.tsx
```diff
@@ -6,36 +6,39 @@ import { TextInput } from '@cityofboston/react-fleet';
...
   handleProceed: (answers: State) => void;
   handleStepBack: () => void;
 }
 
 interface State {
-  dateOfBirth: string;
+  birthDate: string;
 }
 
+// todo: this component will be a datepicker
```
fionawhim, line 24: TBH a next field with a placeholder and maybe even the fancy auto-slash thing would be fine by me.

services-js/registry-certs/client/store/BirthCertificateRequest.ts
```diff
@@ -0,0 +1,44 @@
...
+  parent2FirstName: '',
+  parent2LastName: '',
+};
+
+export default class BirthCertificateRequest {
+  @observable quantity: number = 0;
+  @observable
+  requestInformation: BirthCertificateRequestInformation = initialRequestInformation;
+
+  public setQuantity(newQuantity: number): void {
```
fionawhim, line 30: These setters should be annotated with `@action`. Are you getting warnings in your console?

services-js/registry-certs/client/types.ts
```diff
@@ -17,4 +20,30 @@ export interface DeathCertificateOrder
       LookupDeathCertificateOrder['deathCertificates']['order']
     > {}
 
-export type CertificateType = 'death' | 'birth';
+// Birth-specific
```
fionawhim, line 23: good comments

fionawhim (changes requested):

services-js/registry-certs/client/birth/QuestionsFlow.tsx
jessicamarcus: oh, you're totally right - thx for the catch.

services-js/registry-certs/client/birth/QuestionsFlow.tsx
jessicamarcus: yes! ✨

services-js/registry-certs/client/birth/ReviewRequest.tsx
jessicamarcus: Oh, I accidentally copied that in - it’s a change I was going to propose for fleet (default styling in iOS gives inputs rounded corners). Thanks for catching.

services-js/registry-certs/client/birth/ReviewRequest.tsx
jessicamarcus: Yes ma’am! I have been meaning to write a post on Medium or something that lists all of them (i.e. these are the pseudo elements in a range input, there are the ones in a number input) in one spot...

services-js/registry-certs/client/birth/questions/DateOfBirth.tsx
jessicamarcus: I’m leaving this as-is because it hasn’t yet been decided what sort of field this will be. My rec has been to use `<input type="date" />` with a fallback [month dropdown][day dropdown][year text field]... but I didn’t want to spend any time on a solution that might be scrapped.

services-js/registry-certs/client/store/BirthCertificateRequest.ts
jessicamarcus: Nope! Is there a mobx debugger extension I should install?

jessicamarcus: I was looking at `DeathCertificatesCart`’s `setQuantity` method when I was working on this, and it didn't use `@action`? Would love to sit down Monday so I can understand this better.

services-js/registry-certs/client/birth/questions/DateOfBirth.tsx
fionawhim: 👍

services-js/registry-certs/client/store/BirthCertificateRequest.ts
fionawhim: Yah let's go through this and see.

fionawhim:
@jessicamarcus Looking over Percy, we should talk about the order of the back / next buttons on mobile.

jessicamarcus:
@fionawhim re: button order on mobile - I was thinking of changing the order so ”next” appears first; does this sync with your thoughts? (if so, I’ll get the change into this PR).

fionawhim:
> @fionawhim re: button order on mobile - I was thinking of changing the order so ”next” appears first; does this sync with your thoughts? (if so, I’ll get the change into this PR).

Yeah, I think "Next" should be first.

If you are using the `g` grid layout you can use `g--r` to change the column order to right-to-left.

services-js/registry-certs/client/birth/ReviewRequest.tsx
jessicamarcus: 👍 I guess we might as well drop the component’s own state in favor of `birthCertificateRequest` managing it fully.

fionawhim: Works for me.

fionawhim (approved):
Great! The Mobx bits all look good.

PR #218, opened 2018-12-14
Adjusts allocation code based on order type

fionawhim:
Birth certificates will have a different allocation code from death
certificates. The order.orderType is currently being sent by
registry-certs.


jessicamarcus (approved):
🎉

PR #221, opened 2018-12-21
Rework of error handling in checkout flow

fionawhim:
- Keeps Stripe card errors (e.g. incorrect CVV code) from being logged
   Rollbar
 - Changes to use StatusModal for both in-progress status and error
   messages, so we don’t need a custom screen reader announcement
 - If the order failed because of a credit card error, takes the user
   back to the payment information part of the flow

Removes the "processingError" from Order because it hangs around between
Content pages unless we purposefully remember to clear it out. Changes
tokenization and submission to throw errors instead, which are kept in
State in the Payment and Review Content components, respectively.

Also:
 - Changes the auto-generated types file to be .ts instead of .d.ts so
   that its enums can be imported.
 - Removes quirky and unnecessary re-exporting of CheckoutPage, which I
   think was a holdover from some Flow type checking.

### DIFF
--- a/services-js/registry-certs/client/queries/graphql-types.ts
+++ b/services-js/registry-certs/client/queries/graphql-types.ts
@@ -128,8 +128,18 @@ export interface SearchDeathCertificatesVariables {
 // GraphQL mutation operation: SubmitDeathCertificateOrder
 // ====================================================
 
-export interface SubmitDeathCertificateOrder_submitDeathCertificateOrder {
+export interface SubmitDeathCertificateOrder_submitDeathCertificateOrder_order {
   id: string;
+}
+
+export interface SubmitDeathCertificateOrder_submitDeathCertificateOrder_error {
+  message: string;
+  cause: OrderErrorCause;
+}
+
+export interface SubmitDeathCertificateOrder_submitDeathCertificateOrder {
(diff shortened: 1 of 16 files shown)


services-js/registry-certs/client/queries/graphql-types.ts
```diff
@@ -128,10 +128,20 @@ export interface SearchDeathCertificatesVariables {
 // GraphQL mutation operation: SubmitDeathCertificateOrder
 // ====================================================
 
-export interface SubmitDeathCertificateOrder_submitDeathCertificateOrder {
+export interface SubmitDeathCertificateOrder_submitDeathCertificateOrder_order {
```
jessicamarcus, line 131: Sooo ugly! Thanks a lot, GraphQL. 😆

jessicamarcus (approved):

fionawhim:
@rachelbraun Shoulder-approved the Percy changes.

services-js/registry-certs/client/queries/graphql-types.ts
fionawhim: Yah. This is why sometimes I just slip `type` aliases into `types.ts`.

Issue #228, opened 2019-01-03
Fix single logout


fionawhim:
Logout is not working.

 - [x] Update UI to redirect the user to https://sso-dev.boston.gov/idp/startSLO.ping rather than creating a SAML logout request
 - [ ] Figure out why the logout success URL is 404ing

fionawhim:
Blocked on waiting for LikeMinds to respond about 404 behavior.

fionawhim:
This is now fixed.

PR #229, opened 2019-01-04
Updates logout behavior

fionawhim:
- Redirect to Ping to start single logout process, rather than making a
   SAML request
 - Handle logout requests coming in via POST
 - Pass RelayState when creating logout success responses

See: #228

Also:
 - Fixes incorrect assumption / typecast that payload was a string
   during POST assert processing (Hapi parses the form encoding and
   generates an object)

### DIFF
--- a/services-js/access-boston/src/server/login-auth.ts
+++ b/services-js/access-boston/src/server/login-auth.ts
@@ -1,6 +1,16 @@
-import { Server as HapiServer, RequestQuery } from 'hapi';
+import {
+  Server as HapiServer,
+  RequestQuery,
+  ResponseToolkit,
+  Request as HapiRequest,
+} from 'hapi';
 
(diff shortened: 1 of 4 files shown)


services-js/access-boston/src/server/login-auth.ts
```diff
@@ -202,33 +215,47 @@ export async function addLoginAuth(
     // "required."
     options: { auth: { mode: 'try' } },
     handler: async (request, h) => {
-      const assertResult = await samlAuth.handleGetAssert(
-        request.query as RequestQuery
-      );
+      const query = request.query as RequestQuery;
```
jessicamarcus, line 218: I like the skimmable clarity of this change.

jessicamarcus (approved):
lgtm!

PR #230, opened 2019-01-04
Removes the birth certificate search query

fionawhim:
Since we’re changing to just make requests rather than look up
certificates, we don’t need to query the database this way.


jessicamarcus (approved):
:+1:

Issue #223, opened 2019-01-02
Back button doesn't work in checkout flow


fionawhim:
Ideal back button behavior, per @fionawhim and @jessicamarcus:

 - Clicking "back" takes you to the previous page
 - Data on the previous page should be preserved
 - Clicking back then forward should leave you in the same place
 - Clicking back, modifying form, then clicking forward w/o submitting should not have the form modifications
 - Hacking the URL to bypass steps should show errors on the "review" page w/ missing info (alternate: redirect to first missing place instead?)
 - Reloading a page should preserve submitted data (e.g. on "Review" page) but clear out any changes.

Changes to make: 

 - [x] Change text fields to not write to the `Order` unless submitted
 - [x] Add in URL parameters so Next routes w/ `pushState` instead of `replaceState`
 - [x] Switch order storage from instance variable on `CheckoutPage` to session-backed storage
 - [x] Add error state on review page for missing shipping / billing / empty cart
 - [x] Add tokenized card state to payment page (w/ "clear" link to make CC fields come back)

fionawhim:
Ideal back button behavior, per @fionawhim and @jessicamarcus:

 - Clicking "back" takes you to the previous page
 - Data on the previous page should be preserved
 - Clicking back then forward should leave you in the same place
 - Clicking back, modifying form, then clicking forward w/o submitting should not have the form modifications
 - Hacking the URL to bypass steps should show errors on the "review" page w/ missing info (alternate: redirect to first missing place instead?)
 - Reloading a page should preserve submitted data (e.g. on "Review" page) but clear out any changes.

Changes to make: 

 - [ ] Change text fields to not write to the `Order` unless submitted
 - [ ] Add in URL parameters so Next routes w/ `pushState` instead of `replaceState`
 - [ ] Switch order storage from instance variable on `CheckoutPage` to session-backed storage
 - [ ] Add error state on review page for missing shipping / billing / empty cart
 - [ ] Add tokenized card state to payment page (w/ "clear" link to make CC fields come back)

fionawhim:
Currently dividing and conquering:

 - @jessicamarcus to finish making `PaymentContent` use Formik and not insta-update the `Order` object
 - @fionawhim to have `Order` objects save / restore from session storage

PR #231, opened 2019-01-04
Stores the current Order in session storage

fionawhim:
Using session storage to back the Order means that the user can leave
the CheckoutPage component (by going to the Cart page, for example)
without losing partially filled-out order info.

Moves the card token into OrderInfo so that it’s kept in session
storage.

See #223


services-js/registry-certs/client/store/OrderProvider.ts
```diff
@@ -30,28 +52,67 @@ export default class OrderProvider {
...
+    const { sessionStorage } = this;
+
+    if (sessionStorage) {
+      sessionStorage.removeItem(SESSION_STORAGE_KEY);
+    }
+  }
+
+  /**
+   * Returns a filtered OrderInfo based on the values of storeContactAndShipping
+   * and storeBilling. If they're true, includes the related values, '' if
```
jessicamarcus: This comment is a little unclear; could we go with something along the lines of:

```suggestion
   * and storeBilling. If either is true, include its value, otherwise include an empty string
```

jessicamarcus (approved):
I like this shift in approach for current orders 👍

PR #233, opened 2019-01-04
Adds birth cert submission to CheckoutDao

fionawhim:
- Adds a mutation query function for the client to use
 - Factors out submission response / error handling so that it can be
   shared by death and birth
 - Updates the GraphQL mutation args to take more of the data that’s
   being requested in the question flow


jessicamarcus (approved):
🎉lgtm

PR #234, opened 2019-01-07
Updates Access Boston workflows

### LINKED ISSUE #227: Use new Forgot Password token
Ping is now sending a "userAccessToken" during login assertions for the forgot password endpoint.

We need to save that in the session and send it back as "Token" when triggering the forgot password workflow.

fionawhim:
- Gets the userAccessToken from forgot password login assertions and
   passes it through to the reset password workflow
 - Changes user registration to use a workflow rather that PUTting to
   the user endpoint directly

Fixes #227


jessicamarcus (approved):
lgtm!

PR #235, opened 2019-01-08
Initial port of Public Notices

fionawhim:
Re-implementation of the Vue app found at CityOfBoston/notice-signage

TODO:
 - Auto-reload when the Next app changes
 - Rollbar API key support

Also:
 - Makes the Percy screenshot widths parameterizable


services-js/public-notices/src/components/NoticeDetail.tsx
```diff
@@ -0,0 +1,324 @@
...
+              dangerouslySetInnerHTML={{
+                __html: notice ? notice.title : '&nbsp;',
+              }}
+            />
+          </div>
+
+          <ul className="notice-details">
+            {notice &&
+              (notice.canceled !== '1' ? (
+                <li className="notice-detail">
```
jessicamarcus: Since the markup structure doesn’t change between states, could we do something like this instead?

```jsx
<li className="notice-detail">
  <div className="notice-detail-header">When</div>
  <div className="notice-detail-content">
    {notice.canceled !== '1' ? (
      <>
        <div dangerouslySetInnerHTML={{ __html: notice.notice_date }} />
        <div dangerouslySetInnerHTML={{ __html: notice.notice_time }} />
      </>
    ) : (
      <span className="notice-canceled">Canceled</span>
    )}
  </div>
</li>

File: services-js/public-notices/src/components/NoticeList.tsx [hunk 2]
```diff
@@ -0,0 +1,173 @@
...
+            dangerouslySetInnerHTML={{ __html: notice.title }}
+          />
+
+          <div className="notice-location">
+            {notice.location_street}
+            {notice.location_room && <span>, {notice.location_room}</span>}
+          </div>
+        </div>
+
+        {notice.canceled !== '1' ? (
```
jessicamarcus: Similar comment as in `NoticeDetail.tsx`...

services-js/public-notices/src/model/NoticeController.ts
```diff
@@ -0,0 +1,96 @@
+import { observable, action } from 'mobx';
```
jessicamarcus, line 1: 👍 doc comments!

services-js/public-notices/src/components/NoticeList.tsx
```diff
@@ -0,0 +1,173 @@
...
+    display: block;
+    text-decoration: none;
+  }
+
+  .notice-title a {
+    text-decoration: none;
+  }
+`;
+
+function renderNotice(notice: Notice, label: string | null) {
```
jessicamarcus: Can we add a return type? (here and in the other React components)

jessicamarcus (changes requested):
Lookin’ good! Psyched to see this moving into the monorepo.  🎉

services-js/public-notices/src/components/NoticeDetail.tsx
fionawhim: Done.

services-js/public-notices/src/components/NoticeList.tsx
fionawhim: 👍 all set.

services-js/public-notices/src/components/NoticeList.tsx
fionawhim: Done. `ReactElement<any>` seems to be what `FunctionComponent` is defined to return in the React type definitions.

fionawhim:
@jessicamarcus Ok, all updated!

jessicamarcus (approved):

PR #236, opened 2019-01-09
Makes public notices deployable to S3

fionawhim:
- predeploy script to build and export
 - allows assetPrefix to be set so it can be served from a subpath
 - updates s3-apps-deploy to pull in the config bucket so that .env
   files can be used during the static build


jessicamarcus (approved):
lgtm!

PR #237, opened 2019-01-10
Changes AccessBoston to use workflow error message

fionawhim:
The workflows now give human-understandable error messages, so we add
them as-is.


jessicamarcus (approved):
Human-readable feedback is 💯 !

PR #239, opened 2019-01-11
Changes Public Notices to only show 12 on the side

fionawhim:
Request from Sebastian because 13 can cause some notices to be cut off
when enough locations wrap to other lines.

Tweaks story to show the list w/ height of 1080 and ensure that it has a
canceled item in it.

Also updates spelling from "cancelled" -> "canceled" in a few places.


services-js/public-notices/src/components/NoticeDetail.stories.tsx
```diff
@@ -8,7 +8,7 @@ import NoticeDetail from './NoticeDetail';
 
 const NOTICES: Notice[] = require('../../fixtures/notices.json');
 const LONG_NOTICE = NOTICES[1];
-const CANCELLED_NOTICE = NOTICES[17];
+const CANCELED_NOTICE = NOTICES[17];
```
jessicamarcus, line 11: Why do we gotta hate on the queen’s english?? 😜

services-js/public-notices/src/components/NoticeList.stories.tsx
```diff
@@ -8,9 +8,12 @@ import NoticeList from './NoticeList';
 const NOTICES: Notice[] = require('../../fixtures/notices.json');
 
 storiesOf('NoticeList', module).add('default', () => (
-  <div style={{ width: 450, backgroundColor: 'white' }}>
+  // This width and height match the column when it’s on a TV.
```
jessicamarcus: Good tidbit of explanation on where those values came from. Are they valid for _any_ TV, or one model in particular? If the latter, we should add that addt’l info to the comment.

jessicamarcus (approved):
lgtm!

services-js/public-notices/src/components/NoticeDetail.stories.tsx
fionawhim: colonialsoldierwithprisoner

services-js/public-notices/src/components/NoticeList.stories.tsx
fionawhim: Updated comment to specify 1080p TV.

PR #240, opened 2019-01-11
Updates hapi-next to abandon completed requests

fionawhim:
When getInitialProps does a redirect by writing to the request directly,
this keeps Hapi from trying to process the request further.

Prevents a "Can't set headers after they are sent" error in Access
Boston in cases where session or other middleware is trying to modify
things after the response has been written.

Tested the two places where res.writeHead(...) and res.end() appear in
our Next code:
 - "/" -> "/register" redirect in Access Boston
 - checkout page redirect in registry-certs


jessicamarcus (approved):
lgtm!

PR #241, opened 2019-01-11
Adds polyfills to support Chrome 37

fionawhim:
This is based on the user-agent that the signage appears to be using.


jessicamarcus (approved):
💯

PR #242, opened 2019-01-14
Adds Formik to checkout flow

jessicamarcus:
Moved form field state management from MobX to Formik: instead, MobX state is updated when the user progresses to the next step in the checkout flow.


services-js/registry-certs/client/death/checkout/PaymentContent.tsx
```diff
@@ -113,13 +144,32 @@ export default class PaymentContent extends React.Component<Props, State> {
     }
   };
 
+  // Since formik expects strings for values, we must convert
+  // billingAddressSameAsShippingAddress back to a boolean value.
```
fionawhim, line 149: 👍

services-js/registry-certs/client/death/checkout/PaymentContent.tsx
```diff
@@ -34,31 +40,56 @@ interface Props {
...
   tokenizationError: null | string;
 }
 
+export interface BillingInfo {
+  storeBilling: boolean;
+
+  cardholderName: string;
+  cardLast4: string;
+
+  billingAddressSameAsShippingAddress: boolean | string; // formik needs this value as a string
```
fionawhim: I'd like to talk through this. We should tighten up the types to ensure that you can’t accidentally change `Order`’s `billingAddressSameAsShippingAddress` to a string.

jessicamarcus: I’d love your insight on how best to handle this, because I don’t particularly like this solution. The issue is that `input type="radio"`’s value is a string, so Formik expects the same. The `BillingInfo` interface is used in several places, and I didn’t think it made sense to change the value to a string overall (it makes much more sense as a boolean). I’m not sure that it makes more sense to create a new interface specifically to be used for the form values within Formik?

jessicamarcus:
Changes look good to me, thanks!

fionawhim (approved):
Yay!

PR #248, opened 2019-01-16
registry-certs checkout flow fixes

fionawhim:
- Uses query parameters to move between pages, so back / forward /
   reload works
 - Changes all localStorage and sessionStorage attaching to happen in
   componentDidMount to remove that as a difference between SSR and
   initial client rendering (React does not guarantee that hydrating
   will succeed if the HTML is different) – this fixes reloading the
   shipping page and the button always being disabled even if the form
   is filled out
 - Updates uses of OrderProvider to work with Promises
 - Fixes PaymentContent to allow initialization with "differentAddress"
 - Runs validations and shows errors in Formik forms for initialValues,
   in the case of Order not being valid (also helps for Storybook)

### DIFF
--- a/services-js/registry-certs/client/Storyshots.test.ts
+++ b/services-js/registry-certs/client/Storyshots.test.ts
@@ -2,6 +2,4 @@ import initStoryshots from '@storybook/addon-storyshots';
 
 require('babel-plugin-require-context-hook/register')();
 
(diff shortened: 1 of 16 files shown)


jessicamarcus (approved):
lgtm! Hopefully this’ll do the trick. 😀

PR #249, opened 2019-01-17
Adds jest project support

fionawhim:
With projects, we can run Jest from the root of the repo and have it
test all the separate sub-projects. This is useful for overall code
coverage and IDE integration.

 - Adds a babel.config.js that references that .babelrc files exist
 - Adds a no-op root tsconfig.json
 - Converts all readFile calls to be relative to the file, rather than
   the assumed project root
 - Upgrades registry-certs to Jest 23.6
 - Adds explicit dependencies on jest-environment-* packages to avoid
   the "environment.teardown is not a function" errors (probably related
   to create-react-app in commissions-search)

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,0 +1,10 @@
+const { workspaces = [] } = require('./package.json');
+
+// This is necessary for the Jest babel transformer — when it’s run from the
+// repo’s root – to find each package’s .babelrc files.
+//
+// See: https://github.com/facebook/jest/issues/6053#issuecomment-383632515 And:
+// https://github.com/vikr01/toms-shuttles/blob/dev/babel.config.js
+module.exports = {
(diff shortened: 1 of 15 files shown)


babel.config.js
```diff
@@ -0,0 +1,10 @@
+const { workspaces = [] } = require('./package.json');
+
+// This allows repo Jest running to find the individual .babelrc
```
jessicamarcus: I don’t quite understand this line; should it be `repo running Jest`?

jessicamarcus: (super-helpful comments tho!)

jessicamarcus (approved):
I didn’t know about this functionality in Jest before today. 😃

babel.config.js
fionawhim: Re-wrote the comment, should be clearer now.

PR #250, opened 2019-01-17
Checkout page errors and payment card update

fionawhim:
Due to back button changes (#223 and #248) it is possible to hack
yourself on to the "review" page without filling out the other values.

This tweaks that page to address that scenario. As it’s not a normal
flow case, we don’t have to be particularly graceful about it.

This also adds a state to the payment page for when the card has already
been tokenized.

Also updates the review page to not allow an empty cart submission.


jessicamarcus (approved):
I think this is a solid fix for the issue. 👍

PR #251, opened 2019-01-18
Move of checkout Content pages into common

fionawhim:
We will be generalizing these for birth / marriage. Moving without
making any other changes so that later diffs are smaller, as GitHub is
not great about diffing + file moving.


jessicamarcus (approved):
Totally agree with giving this file move change its own PR.

PR #252, opened 2019-01-18
Adds new component to library

jessicamarcus:
Adds a new progress bar component to react-fleet.


modules-js/react-fleet/src/components/ProgressBar.tsx
```diff
@@ -0,0 +1,87 @@
+import React from 'react';
+
+import { css } from 'emotion';
+
+import { CHARLES_BLUE, OPTIMISTIC_BLUE } from '../utilities/constants';
+
+interface Props {
+  steps: string[];
```
fionawhim: I'm nervous because this requires an a priori list of all the steps and their names.

It would be more flexible for the component to just take the current step # and the total # of steps, and then a single, optional `stepLabel` to display.

I also get nervous when a user-facing label, like the current step name, also gets used as data (by being used to do an `indexOf`).

modules-js/react-fleet/src/components/ProgressBar.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  return (
+    <div className={PROGRESS_BAR_STYLE}>
+      <progress
+        max={totalSteps}
+        aria-valuemin={1}
+        aria-valuemax={totalSteps}
+        aria-valuenow={currentStepNumber}
+        aria-valuetext={progressText}
+        value={props.currentStepCompleted ? currentStepNumber : currentIndex}
```
fionawhim: This is good. The idea being you see the step animate completed while you’re waiting for the next page to load (for example)?

modules-js/react-fleet/src/components/ProgressBar.tsx
```diff
@@ -0,0 +1,87 @@
...
+      >
+        {progressText}
+      </progress>
+
+      {/* hidden from screenreaders to avoid repetition */}
```
fionawhim, line 49: 👍

fionawhim (changes requested):

modules-js/react-fleet/src/components/ProgressBar.tsx
jessicamarcus: Sure, that sort of thing - or some other thing I haven't thought of yet 😄

modules-js/react-fleet/src/components/ProgressBar.tsx
jessicamarcus: I _personally_ don’t mind using an array of strings like this, but can certainly understand why someone would prefer to avoid. I’ll make that change, and change the `showStepName` property so it accepts a string for the task’s name instead.

fionawhim: Thanks!

fionawhim (approved):
Thanks for making the change! I think numbers-only is cleaner and more flexible.

PR #254, opened 2019-01-22
Requires env var to show birth certs UI in prod

fionawhim:
Keeps us from making the birth cert request UI available on
registry.boston.gov.


jessicamarcus (approved):
Thank you - this makes me feel more confident as we work on birth certs!

PR #256, opened 2019-01-23
Adds a grace period for seeing a deployment

fionawhim:
We're seeing that new deployments aren’t appearing in the service
response right away. This gives us a bit of time to wait for them to
show up before failing the deploy.


jessicamarcus (approved):
💯

PR #258, opened 2019-01-24
Sends birth cert requests to the registry DB

fionawhim:
Uses the new sp_AddBirthRequest stored procedure for birth certificate
requests. Changes addItem to be specifically about death certificates.

Modifies payment-webhooks to be tolerant of uncaptured charges.

Also:
 - Switches to use @types/stripe in order to correctly type our Stripe
   calls and data
 - Changes some fields from protected to private so we get accurate
   unusedness checks on them
 - Tweaks Jest fakes in mutation.test.ts to use Jest’s auto-mocking
   rather than hand-made fakes

### DIFF
--- a/services-js/payment-webhooks/server/stripe-events.ts
+++ b/services-js/payment-webhooks/server/stripe-events.ts
@@ -1,6 +1,6 @@
 /* eslint no-console: 0 */
 
-import { NodeStripe, Event, Charge } from 'stripe';
+import Stripe, { webhooks, charges } from 'stripe';
 import INovah from './services/INovah';
 
 import Rollbar from 'rollbar';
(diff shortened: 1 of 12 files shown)


services-js/payment-webhooks/server/stripe-events.ts
```diff
@@ -1,20 +1,27 @@
...
-  stripe: NodeStripe;
+  stripe: Stripe;
 };
 
 async function processChargeSucceeded(
   { rollbar, stripe, inovah }: Dependencies,
-  charge: Charge
+  charge: charges.ICharge
 ): Promise<void> {
+  // Birth cert charges aren't captured at first, we wait until fulfillment to
```
jessicamarcus: Would it make sense to expand on this comment to make clear that fulfillment is a human-triggered event?

jessicamarcus (approved):
It’s gettin’ real in here! 😆

services-js/payment-webhooks/server/stripe-events.ts
fionawhim: Updated.

jessicamarcus: awesome, thanks!

PR #260, opened 2019-01-24
Error page for react-fleet

fionawhim:
Makes a Boston-branded error / 404 page for commissions-app and
registry-certs so that when they’re down tonight we will have a nicer
message than "500".


jessicamarcus (approved):
Awesome addition to our library!

PR #261, opened 2019-01-25
Makes Access Boston run without IP information

fionawhim:
Allows the server to start up to generate the metadata XML from its
service provider data.

Also adds README info about generating keys and metadata.


jessicamarcus (approved):
lgtm! Nice comments, too.

Issue #259, opened 2019-01-24
Add an identification verification step to the questions flow


jessicamarcus:
The `parentalInformation` step includes the question ”were your/their parents married at the time of your/their birth?”. Any answer other than **yes** should take the user to an id verification step, otherwise the question flow ends and the user proceeds to `/birth/review`.

The actual ID verification functionality will be a discrete component, since there is also a freestanding id verification page to support cases where a user submits a request, and is later asked for ID verification by the Registry.

**user-facing:**
- explanatory content
- Upload ID => proceed to id scan => success => proceed to `/birth/review`
- No ID? => provide messaging and “Request help” button, then end request flow

**requirements:**
- [x] add a step to the overall progress if id verification is required
- [x] addt'l logic for back/next buttons

PR #263, opened 2019-01-25
Renames and reorganizes files; adds id verify step to question flow; …

jessicamarcus:
Lots of little changes; this is more of an intermediate commit in some places, but I wanted to get the changes in sooner rather than later...

- Rename `QuestionFlow` to `QuestionPage` and `ReviewRequest` to `ReviewRequestPage` for better clarity
- Moves styling file up out of `/questions`
- Renames some css style vars
- Moves icon components to their own directory
- Adds `VerifyInformation` placeholder component; this will be the upload/scan functionality component
- Adds `VerifyInformationPage` - when the Registry requests more information from a user at a later time, this is where the user will land
- Adds `VerifyInformation` step to question flow; necessitated moving the “steps” array into `BirthCertificateRequest.ts`, because that step is added (or removed) during the user’s journey
- Adds a birth `PageWrapper` component


jessicamarcus:
This pr fixes https://github.com/CityOfBoston/digital/issues/259

services-js/registry-certs/client/birth/BirthCertificateRequestHeader.tsx
```diff
@@ -1,14 +1,12 @@
...
-
-import { STEPS } from './constants';
+import { ProgressBar } from '@cityofboston/react-fleet';
 
 interface Props {
+  totalSteps: number;
   currentStep: number;
-  stepComplete?: boolean;
+  currentStepCompleted?: boolean;
+  showProgress?: boolean;
```
fionawhim: One note about this is that the step information is required even if `showProgress` is false.

Some alternatives would be to make `Props` a union type (see how `PageLayout.tsx` treats `cart`) or to have an optional `progress` prop that's either `null` or an object w/ `totalSteps`, `currentStep`, `currentStepCompleted`.

services-js/registry-certs/client/birth/QuestionsPage.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+    // If ID verification is not known to be necessary, advance to the
+    // review page.
+    // Otherwise, continue to verifyIdentification step. When completed,
+    // then advance to the review page.
+    if (
+      question === 'parentalInformation' &&
+      !steps.includes('verifyIdentification')
+    ) {
+      Router.push('/birth/review');
```
fionawhim, line 59: I noticed today that these will need a `window.scrollTo(0, 0)` but you can add that later.

services-js/registry-certs/client/birth/QuestionsPage.tsx
```diff
@@ -0,0 +1,158 @@
...
+    setCurrentStep(steps[newIndex]);
+  };
+
+  // Clear all data and return to initial question.
+  private handleUserReset = (): void => {
+    this.props.birthCertificateRequest.clearBirthCertificateRequest();
+  };
+
+  public render() {
+    const answers = this.props.birthCertificateRequest.requestInformation;
```
fionawhim: could also do this as:

`answers: requestInformation` in the object destructuring below. (I think that’s the right order)

services-js/registry-certs/client/birth/QuestionsPage.tsx
```diff
@@ -0,0 +1,158 @@
...
+        <Head>
+          <title>Boston.gov — Request a Birth Certificate</title>
+        </Head>
+
+        {currentStep === 'forWhom' && (
+          <ForWhom
+            forSelf={answers.forSelf}
+            howRelated={answers.howRelated}
+            handleStepCompletion={setCurrentStepCompleted}
+            handleProceed={a => this.answerQuestion('forWhom', a as any)}
```
fionawhim, line 112: double check on whether these still need the `as any` because I see that some don’t have it

services-js/registry-certs/client/birth/QuestionsPage.tsx
```diff
@@ -0,0 +1,158 @@
...
+              this.props.birthCertificateRequest.verificationStepRequired
+            }
+          />
+        )}
+
+        {currentStep === 'verifyIdentification' && (
+          <VerifyIdentification
+            handleStepCompletion={setCurrentStepCompleted}
+            handleProceed={a =>
+              this.answerQuestion('verifyIdentification', a as any)
```
fionawhim, line 152: `as any`

services-js/registry-certs/client/birth/RadioItemComponent.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+  );
+}
+
+interface Props {}
+
+interface State {
+  answer: string;
+}
+
+class SampleQuestion extends React.Component<Props, State> {
```
fionawhim, line 32: If this bit is a pain to write / maintain you could also just get its coverage from the other pages and have these stories just focus on a singular instance.

services-js/registry-certs/client/birth/questions/ParentalInformation.tsx
```diff
@@ -70,76 +71,95 @@ export default class ParentalInformation extends React.Component<Props, State> {
...
-      return this.props.handleStepBack;
-    } else {
+    if (this.mayBeRestricted()) {
       return () => this.setState({ parentsMarried: '' });
+    } else {
+      return this.props.handleStepBack;
+    }
+  }
+
+  public componentDidUpdate(
```
fionawhim, line 106: This surprises me. It seems like the business logic of whether there’s a verification step is in an obscure place. It’s a bit Rube Goldberg to making sure that `verificationStepRequired` is called with the right value.

I'd recommend making a method possibly on `BirthCertificateRequest` that you could feed either `this.state.parentsMarried` or `requestInformation.parentsMarried` into and get a true / false value.

MobX `@calculated` properties may be your friend; I realize there’s also the slightly tricky bit that you’re avoiding updating the global `BirthCertificateRequest` until the form is submitted, but you want the local render to reflect the component’s state.

Let’s chat about this.

services-js/registry-certs/client/birth/questions/ParentalInformation.tsx
```diff
@@ -202,10 +224,14 @@ export default class ParentalInformation extends React.Component<Props, State> {
     );
   }
 
-  renderRestrictedText(): React.ReactChild {
+  private renderRestrictedText(): React.ReactChild {
+    window.scrollTo({ top: 0, left: 0, behavior: 'smooth' });
```
fionawhim: `scrollTo` in a render method isn’t ideal, since they’re supposed to be functions without side effects.

services-js/registry-certs/client/birth/questions/YesNoUnsureComponent.tsx
```diff
@@ -20,47 +19,36 @@ interface Props {
  */
 export default function YesNoUnsureComponent(props: Props): JSX.Element {
   function radioItem(
-    value: YesNoUnknownAnswer,
-    text: string
+    itemValue: YesNoUnknownAnswer,
+    itemText: string
   ): React.ReactChild {
+    // map values to the icons to use
+    const iconName =
```
fionawhim, line 26: One of the things I miss about CoffeeScript is how all of its statements had values, so you could easily use a `switch` in this scenario. Doing one level of ternary operator nesting is fine though, if it means you can keep the `const`.

services-js/registry-certs/client/store/BirthCertificateRequest.ts
```diff
@@ -53,6 +71,29 @@ export default class BirthCertificateRequest {
     this.currentStepCompleted = isCompleted;
   };
 
+  @action
+  public addVerificationStep = (): void => {
```
fionawhim, line 75: Ah, ok.

I’d recommend replacing all of this with a `@computed` `steps` property that would return either `INITIAL_STEPS` or `INITIAL_STEPS_WITH_VERIFICATION`.

services-js/registry-certs/client/store/BirthCertificateRequest.ts
```diff
@@ -53,6 +71,29 @@ export default class BirthCertificateRequest {
...
+    const targetIndex = this.steps.indexOf('parentalInformation') + 1;
+
+    if (!this.steps.includes('verifyIdentification')) {
+      this.steps.splice(targetIndex, 0, 'verifyIdentification');
```
fionawhim, line 81: Warning here that you’re mutating `this.steps`, which was initialized with a global constant. This means that these changes are affecting any future `BirthCertificateRequest` objects that are created.

It might not be a bug here in practice, but it is one of those things that could surprise you later.

jessicamarcus: I started off with `this.steps = this.steps.splice(etc)` but found that when the `verify` step was added to the array, the progress bar component lost the information... I should have left a comment! 😞

services-js/registry-certs/client/store/BirthCertificateRequest.ts
jessicamarcus: Between this and your earlier comment, I definitely want to learn about `@computed`!

services-js/registry-certs/client/birth/RadioItemComponent.stories.tsx
jessicamarcus: Ah, I suppose that’s true!

services-js/registry-certs/client/birth/QuestionsPage.tsx
jessicamarcus: The `forWhom` question is a weird case, because the value is handled as a string inside the component (because Formik), but is a boolean in the request state.

services-js/registry-certs/client/birth/QuestionsPage.tsx
jessicamarcus: Oh right! fyi the order is actually `requestInformation: answers`

services-js/registry-certs/client/birth/QuestionsPage.tsx
jessicamarcus: It needs it here for now (because we haven’t worked out details for uploaded files yet), but I’ve added a todo comment.

services-js/registry-certs/client/birth/RadioItemComponent.stories.tsx
jessicamarcus: But I’m going to leave this in place for now; in order to see the selected and not-selected states, there needs to be a state object _somewhere_ to keep track, and all usages of this component store state in a parent question component.

fionawhim: Isn’t selected passed in as a prop?

jessicamarcus: Not sure I understand what you mean; passed into where?

The `RadioItemComponent` determines for itself whether it’s checked by comparing the current value of the form item with its own value; lmk if the comments don’t explain well and I’ll rewrite.

fionawhim: I was thinking you could show the selected version by having your story pass `questionValue` and `itemValue` props that were the same string literal, and the deselected version by making those props different. Is that not the case?

fionawhim: (Sorry, I only just now went and looked up the exact props to see that you were doing two values rather than a boolean `selected` prop)

fionawhim (approved):
Approving with the note that we'll come back to the verification warning step behavior / implementation next week.

PR #267, opened 2019-01-25
Extends icon component

jessicamarcus:
Since both the id and birth cert icons use the same frame and lines, I combined them into one component.


fionawhim (approved):
👣

PR #275, opened 2019-01-28
Removes punctuation from phone numbers

fionawhim:
Prevents us from confusing Ping with silly characters.

Also adds Array.find polyfill for IE11.


jessicamarcus (approved):
💯

PR #276, opened 2019-01-28
Switches all buttons to use the btn--br variant

fionawhim:
Makes all registry-certs buttons consistently have the new bordered
button style from Fleet.

Also removes button row styling in favor of Fleet grids:
 - Uses g--r to put "Next" first in source order but to the right of
   the screen
 - Uses small columns and btn--b to make buttons take up an ok amount of
   space at large breakpoints but the entire column when the grid
   collapses at low breakpoints
 - Throws in some margin classes


fionawhim:
One future enhancement I think is to have a class that turns an element to a block/width100% at low breakpoints, so the buttons can be naturally sized when they're on the same row.

jessicamarcus (approved):
👍

PR #282, opened 2019-01-30
Updates to device registration flow

fionawhim:
- Hides all but last 2 digits of phone numbers, per security team
   request
 - Prevents MFA registration if the user account is not marked as
   needing registration (limits unauthorized adding of MFA devices)
 - Switches to use a <form> in code input so we get return key handling
   for free


services-js/access-boston/src/client/device-verification/DeviceVerificationModal.tsx
```diff
@@ -207,7 +214,16 @@ export default class DeviceVerificationModal extends React.Component<Props> {
             try a different number or email
           </button>.
         </div>
-      </>
+      </form>
     );
   };
 }
+
+function elidePhoneNumber(phoneNumber: string | null): string {
```
jessicamarcus, line 222: I feel like this could be a useful function to keep _some_ where for future use... (gitbook? react-fleet? an as-of-yet nonexistent `cob-utility-functions` module?)

jessicamarcus (approved):
lgtm!

services-js/access-boston/src/client/device-verification/DeviceVerificationModal.tsx
fionawhim: yep, we should factor it out next time we need it.

PR #283, opened 2019-01-30
Access Boston review punchlist

fionawhim:
- Removes group name parsing because the format changed
 - Adds cancel link at the bottom of change password
 - Allows apps to be specified as "open in new window"


jessicamarcus (approved):
👍

PR #285, opened 2019-01-31
Access Boston apps updates

fionawhim:
- Modifies / moves AppRegistry to work in the browser for Storybook
 - Changes index page storybook to use the fixture YAML file
 - Updates Jest / Storybook to allow for YAML importing
 - Moves app images to the static directory that is actually hosted,
   also adds a few more
 - Tweaks app image styles to work better at small breakpoints without
   distorting or being hella tall


jessicamarcus (approved):
lgtm!

PR #286, opened 2019-02-01
Adds static dir when building Storybook for Percy

fionawhim:
Will allow the app images to appear in the images.


jessicamarcus (approved):
👍

PR #288, opened 2019-02-01
Adds date input component to library

jessicamarcus:
New component for react-fleet; allows a user to supply a memorable date, such as a birthdate.


modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.test.ts
```diff
@@ -0,0 +1,188 @@
+import { isDateValid, setDateLimit } from './MemorableDateInput';
+
+const today = new Date(new Date().setHours(0, 0, 0, 0));
```
fionawhim: Lock this in to being a specific date, otherwise you put a timebomb in the code when tests start failing next year.

You can use the 'mockdate' package to do so. See death-certificates.test.ts

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.test.ts
```diff
@@ -0,0 +1,188 @@
...
+
+describe('set upper limit', () => {
+  const latestDate = '12/12/2020';
+
+  test('latestDate provided', () => {
+    expect(setDateLimit(latestDate)).toEqual(new Date(latestDate));
+  });
+
+  test('onlyAllowPast', () => {
+    expect(setDateLimit('', true)).toEqual(today);
```
fionawhim: I'm not understanding how this test is different from the "onlyAllowFuture" test above.

fionawhim: Sometimes using `it` instead of `test` can make things clearer because you can make a little sentence to label the test.

```
describe('setDateLimit', () => {
  it('returns the current date if the something something', () => {
  });
});
```

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
```diff
@@ -0,0 +1,458 @@
...
+  if (isNotValid && isNotValid.length > 0) {
+    throw new Error(
+      `Initial date is not valid: ${isNotValid}
+    Received: initialDate: ${props.initialDate}`
+    );
+  }
+}
+
+// Determine upper or lower limit date, if applicable.
+export function setDateLimit(
```
fionawhim: I'm feeling a mismatch between the name `setDateLimit` and what this function is doing. It's more like `stringToDate` but with the `hasRelativeLimitation` conditional.

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
```diff
@@ -0,0 +1,458 @@
...
+    date = today;
+  }
+
+  return date;
+}
+
+/**
+ * Returns an error string if user input is incomplete.
+ */
+export function isInputComplete(fields): string {
```
fionawhim, line 350: It would be great to have a type on `fields`

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
```diff
@@ -0,0 +1,458 @@
...
+}
+
+/**
+ * Returns useful error string if a given date is not valid.
+ */
+export function isDateValid(
+  date: Date,
+  earliest: Date | null,
+  latest: Date | null,
+  onlyAllowPast: boolean | undefined,
```
fionawhim: You could do `onlyAllowPast: boolean = false` if you wanted, to avoid `undefined`s.

fionawhim: Also, when the number of options gets this high, consider putting them in an options Object that gets passed. That way you don't need to remember / look up if the params go "onlyAllowPast, onlyAllowFuture" or "onlyAllowFuture, onlyAllowPast" or pad out will `null` date arguments. It basically ups the documentation level a bit at the call site.

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.test.ts
```diff
@@ -0,0 +1,188 @@
...
+  });
+});
+
+describe('isDateValid', () => {
+  const pastDate = '12/12/2010';
+  const middleDate = '6/16/2016';
+  const upcomingDate = '9/7/2020';
+  const futureDate = '1/1/2030';
+
+  test(`onlyAllowPast | user entered: ${pastDate}`, () => {
```
fionawhim: You can nest `describe`s as well.

E.g.:

```
describe('isDateValid', () => {
  describe('onlyAllowPast', () => {
    it('does not error on a date in the past', …);
    it('errors on a date in the future', …);
    it('respects an "earliest" parameter', …);
  });
});

File: modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.test.ts [hunk 7]
```diff
@@ -0,0 +1,188 @@
...
+
+  test(`no limits specified | user entered: ${upcomingDate} then ${pastDate}`, () => {
+    expect(
+      isDateValid(new Date(upcomingDate), null, null, undefined, undefined)
+    ).toEqual('');
+
+    expect(
+      isDateValid(new Date(pastDate), null, null, undefined, undefined)
+    ).toEqual('');
+  });
```
fionawhim, line 247: I notice you're missing coverage for `onlyAllowFuture` or `onlyAllowPast` without any other dates specified.

(TBH you could consider dropping the `onlyAllowFuture`/`onlyAllowPast` arguments for this function in favor of just the `earliest`/`latest` `Date`s. I know that `onlyAllowFuture`/`onlyAllowPast` might be good conveniences for the component’s Props, but here I think their only value is tweaking the error message. It would simplify things.)

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
```diff
@@ -0,0 +1,458 @@
+import React from 'react';
+
+import { css } from 'emotion';
+
+interface Props {
+  initialDate?: string;
```
fionawhim: should this be a `string` or a `Date`? I think the latter for symmetry with this component’s output.

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
```diff
@@ -0,0 +1,458 @@
...
+ * An initial date can be passed in via the initialDate property.
+ *
+ * onlyAllowPast: If true, will only accept yesterday or earlier for a date.
+ * onlyAllowFuture: If true, will only accept tomorrow or later for a date.
+ *
+ * Whenever the date is updated, this.props.handleDate is called.
+ *
+ * *Please note that all dates are relative to the user’s computer.
+ *
+ * Inspired by https://design-system.service.gov.uk/components/date-input/
```
fionawhim, line 48: Good comment, esp. this link.

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
```diff
@@ -0,0 +1,458 @@
...
+      error: '',
+      date: this.initial,
+      fields: this.initial
+        ? {
+            month: this.initial.getMonth(),
+            day: this.initial.getDate(),
+            year: this.initial.getFullYear(),
+          }
+        : ({
+            month: '',
```
fionawhim, line 92: You should settle on an accurate type of the `fields`. I'd recommend `number | null` but you could do `string` instead.

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
```diff
@@ -0,0 +1,458 @@
...
+
+    this.initial = props.initialDate ? new Date(props.initialDate) : null;
+
+    checkIfPropsValid(props);
+
+    this.earliest = setDateLimit(
+      props.earliestDate || '',
+      !!props.onlyAllowFuture
+    );
+    this.latest = setDateLimit(props.latestDate || '', !!props.onlyAllowPast);
```
fionawhim: Do you actually get value out of letting `this.earliest`/`this.latest` to `today` if the `onlyAllow` props are given?

Based on my reading of `isDateValid`, the earliest/latest dates aren't considered if `onlyAllowFuture`/`onlyAllowPast` are set (respectively).

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
```diff
@@ -0,0 +1,458 @@
...
+    );
+
+    this.setState({ error: errorString });
+
+    return errorString.length === 0;
+  }
+
+  // If date is within range, update value. Otherwise, clear the value.
+  private updateDate(): void {
+    let date: Date | null = new Date(
```
fionawhim: Given that it’s unclear what locale is being used for this parsing, it might be better to make the `Date` using the `(year, month [, date [, hours [, minutes [, seconds [, ms ] ] ] ] ] )` arguments. (You have numbers anyway.)

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
```diff
@@ -0,0 +1,458 @@
...
+      }`
+    );
+
+    // If supplied date is out of range, clear the date value.
+    if (!this.checkIfDateInRange(date)) {
+      date = null;
+    }
+
+    // Update state and call parent component’s method with the new date value.
+    this.setState({ date }, () => {
```
fionawhim: It looks like you don’t actually need to keep the date in `State`, you can just use `fields` and call `handleDate`

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
```diff
@@ -0,0 +1,458 @@
...
+  private handleBlur = (): void => {
+    this.timeout = setTimeout(() => this.setState({ hasFocus: false }), 0);
+  };
+
+  // Handler for the user-controlled input fields.
+  private handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    this.setState({
+      fields: {
+        ...this.state.fields,
+        [event.target.name]: event.target.value,
```
fionawhim, line 148: Be aware that you’re setting `fields` to `string`s here, not `number`s.

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
```diff
@@ -0,0 +1,458 @@
...
+    }
+
+    // Update state and call parent component’s method with the new date value.
+    this.setState({ date }, () => {
+      this.props.handleDate(date);
+    });
+  }
+
+  // Focus and blur handlers are used to determine whether a user is focused
+  // on any of the fields for this component.
```
fionawhim: You should make it clearer here that the goal is to track focus / blur across all 3 fields, which is why you need the `setTimeout`.

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
```diff
@@ -0,0 +1,458 @@
...
+  } else if (missingFields.length === 2) {
+    errorString = `Date must include the ${missingFields[0]} and ${
+      missingFields[1]
+    }`;
+    // All fields missing
+  } else if (missingFields.length === 3) {
+    errorString = 'Please enter a date';
+  }
+
+  return errorString;
```
fionawhim, line 381: This function would be an ideal one to have tests for (explicitly checking that the output says "Date must include the day and year" or whatevs).

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
```diff
@@ -0,0 +1,458 @@
...
+  ): void {
+    const { fields } = this.state;
+    const errorString = isInputComplete(fields);
+
+    // Only calculate a Date from the user-supplied values when a complete date
+    // has been entered by the user.
+    if (prevState.fields !== fields && errorString.length === 0) {
+      this.updateDate();
+    }
+
```
fionawhim, line 165: I'm coming around a little bit on your use of `componentDidUpdate`. I think that there’s value in using it to look at `fields`, since it does mean that you can call `updateDate` in response to changes in to `fields` regardless of how they occurred (whether in `handleChange` or any other methods).

Just be mindful that you’re not writing imperative / mutation code that could be better handled as a function in `render`. For example, does `error` have to be in `State`? Or could you just calculate it in `render` from the current value of `fields` and `hasFocus`? (Or some State instead of `hasFocus`, like `touched`.)

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.test.ts
jessicamarcus: I go back and forth between preferring `it` and `test`... I’ll make some updates.

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
jessicamarcus: How do you feel about `returnLimitAsDate`?

jessicamarcus: or `returnDateLimit` I suppose; “set” definitely is not a good keyword here.

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
jessicamarcus: I’ll expand on the comment.

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
jessicamarcus: I wanted to accept strings for `initialDate`, `earliestDate`, and `latestDate` because `<input type="date">` accepts strings for its correlating attributes `value`, `min`, and `max` and it seems appropriate to follow a similar pattern. Now that I’m talking about it “out loud” though, I wonder if it’s better to have the property names match those of the native input?

fionawhim: The way I'm thinking of it is that this component can be used to give me `Date` objects (via the `handleDate` callback), which I’ll then store in various model objects.

When it comes time to edit one of my models, I’ll want to pass its `Date` back into the editor to display.

That internal consistency feels more important to me than matching an HTML spec.

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
jessicamarcus: If `null` is used, React throws
```console
Warning: `value` prop on `input` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.
```
`undefined` causes it to treat the fields like uncontrolled components and it gets crankier. 😜

But for sure, I’ll add the overlooked `Fields` type!

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
jessicamarcus: I can see that. I’ll change the component to accept `string | Date` and allow for either case.

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
jessicamarcus: If `this.earliest`/`this.latest` are present, they determine the `min` and `max` values for the year number field.

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
jessicamarcus: Ah, thanks for the catch!

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
fionawhim: In the logic of your component, the fields should probably be `number`s. You’ll need to translate that into `string`s when rendering and back to `number`s when the `value` changes, since `<input>`’s `value` attribute is a `string`.

Or, just keep them in your `State` as `string`s and translate when you’re constructing the `Date`.

Whichever you choose isn’t as important as being clear about the types.

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.test.ts
```diff
@@ -0,0 +1,220 @@
...
+  const dateFromPast = 'The date must be from the past';
+  const dateFromFuture = 'The date must be in the future';
+
+  it(`passes if user entered ${pastDate} and onlyAllowPast is true`, () => {
+    expect(
+      isDateValid(new Date(pastDate), null, TODAY, true, undefined)
+    ).toEqual('');
+  });
+
+  it(`returns “${dateFromPast}” if user entered ${futureDate} and onlyAllowPast is true`, () => {
```
fionawhim: 👍

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.test.ts
```diff
@@ -0,0 +1,220 @@
...
+    ).toEqual('');
+  });
+
+  it(`returns “${dateFromFuture}” if user entered ${pastDate} and onlyAllowFuture is true`, () => {
+    expect(
+      isDateValid(new Date(pastDate), TODAY, null, undefined, true)
+    ).toEqual(dateFromFuture);
+  });
+
+  it(`passes if user entered ${pastDate}, earliestDate is ${middleDate}, and onlyAllowPast is true`, () => {
```
fionawhim: This test should have exposed a bug in your code.

Using the constants for the dates in your test is making things hard to understand. I'm having to keep scrolling back and forth. I think that if this file said instead:

"It passes if the user entered 12/12/2010, the earliest date is 6/16/2016, and onlyAllowPast is true"

you would have noticed that that's not what you want.

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.test.ts
```diff
@@ -0,0 +1,220 @@
...
+  it('should set the date limit to latestDate', () => {
+    expect(returnLimitAsDate(latestDate)).toEqual(new Date(latestDate));
+  });
+
+  it('returns today’s date if onlyAllowFuture or onlyAllowPast is specified', () => {
+    expect(returnLimitAsDate('', true)).toEqual(TODAY);
+  });
+});
+
+describe('isDateValid', () => {
```
fionawhim, line 36: I just want to highlight that there are a lot of cases to test here since you need to cover how the boolean arguments interact with the date range arguments. That might be worth it to get the error messages that you want.

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
```diff
@@ -0,0 +1,468 @@
...
+    this.setState({ error: errorString });
+
+    return errorString.length === 0;
+  }
+
+  // If date is within range, update value. Otherwise, clear the value.
+  private updateDate(): void {
+    let date: Date | null = new Date(
+      new Date(
```
fionawhim: Where did we land on how this works if the browser is in other locales? Can you use the constructor that takes numbers instead so it’s not a concern?

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.test.ts
```diff
@@ -0,0 +1,220 @@
...
+    ).toEqual('');
+
+    expect(
+      isDateValid(new Date(pastDate), null, null, undefined, undefined)
+    ).toEqual('');
+  });
+
+  describe('isInputComplete', () => {
+    const missingDayYear = 'Date must include the day and year';
+    const missingMonth = 'Date must include the month';
```
fionawhim, line 229: Nice.

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
```diff
@@ -0,0 +1,468 @@
...
+          </div>
+
+          <div>
+            <label htmlFor="year" className="txt-l">
+              Year
+            </label>
+
+            <input
+              {...inputAttributes}
+              id="year"
```
fionawhim: I'm just noticing now that these IDs are in danger of colliding if you put more than one date picker on a page. Could save that for a future bug fix, though.

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.test.ts
```diff
@@ -0,0 +1,220 @@
...
+import {
+  isDateValid,
+  isInputComplete,
+  returnLimitAsDate,
+} from './MemorableDateInput';
+
+const TODAY = new Date('2/1/2016');
+
+beforeEach(() => {
+  MockDate.set(TODAY);
```
fionawhim, line 12: Don't forget about making this be some time of day other than the default, to shake out any bugs around when you set hours/minutes/seconds to 0.

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
```diff
@@ -0,0 +1,468 @@
...
+): string {
+  const dateTime = date.getTime();
+  const todayTime = new Date(new Date().setHours(0, 0, 0, 0)).getTime();
+
+  let errorString = '';
+
+  if (onlyAllowPast) {
+    if (dateTime > todayTime) {
+      errorString = 'The date must be from the past';
+    } else if (earliest && dateTime > earliest.getTime()) {
```
fionawhim: Ah, ok, here is the bug I think. This should be `<` since you don’t want `dateTime` to be less than the earliest date.

fionawhim (changes requested):

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
jessicamarcus: I _hate_ it when I overlook this... thanks for catching.

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
```diff
@@ -0,0 +1,476 @@
...
+
+    return errorString.length === 0;
+  }
+
+  // If date is within range, update value. Otherwise, clear the value.
+  private updateDate(): void {
+    const { year, month, day } = this.state.fields;
+
+    // When Date is called with more than one arg, result is set to midnight.
+    let date: Date | null = new Date(+year, +month - 1, +day);
```
fionawhim, line 119: 👍

fionawhim (approved):

PR #289, opened 2019-02-04
Persistent query API for registry-certs backend

fionawhim:
- Adds a new helper to support persistent queries to hapi-common
 - Updates header keys authentication to allow extra data for the
   credentials object
 - Adds chargeOrder mutation and a charge-bc-order-v1 persistent query
   that the fulfillment backend can use, that requires that it’s run
   by the fulfillment API key


fionawhim:
re-running test. My last-minute rename didn't all-the-way take.

jessicamarcus (approved):
lgtm!

PR #294, opened 2019-02-05
Puts first / last name in header

fionawhim:
Also removes "Request Access" in account tools from apps.yaml fixture.
(It has already been removed from the dev / test configs.)

Fixes CityOfBoston/iam#383
Fixes CityOfBoston/iam#382


jessicamarcus (approved):
👍

PR #296, opened 2019-02-06
Upload file / delete file APIs for birth certs

fionawhim:
Upload API is multipart since that's the most straightforward for
browsers. Delete is GraphQL mutation.

Also adds mime-types to deploy-tools, since it was being imported
without an explicit dependency.


jessicamarcus (approved):
lgtm!

PR #298, opened 2019-02-06
Babel compilation / polyfill overhaul

fionawhim:
Takes guesswork out of polyfilling by applying standard polyfills needed
by our common UI libraries, and having Babel automatically put in
polyfills for other ES6 functions we happen to use.

Also moves a lot of copied babelrc compliation for handling Jest /
Storyshots and puts it into next-client-common.


modules-js/next-client-common/babel.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * Babel preset to bring in next/babel and set its polyfill policy to "usage".
```
jessicamarcus, line 2: good explanation

services-js/public-notices/src/pages/_document.tsx
```diff
@@ -52,10 +52,6 @@ export default class extends Document {
       <html lang="en" className="js flexbox">
         <Head>
           <meta name="viewport" content="width=device-width, initial-scale=1" />
-          <script
```
jessicamarcus, line 55: Removing inline `<script>` tags always makes me so happy...

jessicamarcus (approved):
💯

PR #304, opened 2019-02-08
Email updates to prep for more emails

fionawhim:
- Renames "Receipt Email" to "Death Receipt Email"
 - Changes pattern to have all templates loaded into functions in an
   EmailTemplates object, rather than separate model objects for each
   email type
 - Makes subject be a template as well


fionawhim:
verified on staging that emails still going out

jessicamarcus (approved):
💯

PR #305, opened 2019-02-12
Updates password / device flags in registration

fionawhim:
Keeps stages of the registration flow from repeating if the user clicks
around during the process.

See CityOfBoston/iam#405


jessicamarcus (approved):
lgtm!

PR #307, opened 2019-02-12
Adds UploadPhoto component, etc. - resolves #264

### LINKED ISSUE #264: UI for ID upload step in questions flow
Component to enable a user to upload images and documents for identification verification. 

This component will be used in the questions flow `verifyInformation` step, as well as a freestanding page; when the Registry manually requests id verification from a user for an existing order, this page is where that user is directed.

**Waiting on:**
- [ ] detailed instructions for the user
- [x] visual design for interface
- [x] allowed file types for documents

**Functional requirements:**
- user must submit front and back images of their id
- user can add addt’l supporting documents
- user can review their images before submitting
- user can use their device’s camera to take a photo of documents, if possible

jessicamarcus:
- Adds new component to react-fleet
- Copies and adapts existing file input to accept multiple files
- Expands on `VerifyInformationComponent`’s functionality
- Updates Verify Identification in questions
- Changes usages of `window.scrollTo` to `window.scroll`


modules-js/react-fleet/package.json
```diff
@@ -24,6 +24,8 @@
     "preset": "@cityofboston/config-jest-babel"
   },
   "dependencies": {
+    "react-dropzone": "^8.1.0",
+    "@cityofboston/next-client-common": "^0.0.0"
```
fionawhim: I don't see an instance of next-client-common being used by react-fleet. Am I missing something?

modules-js/react-fleet/src/form-elements/UploadPhoto.tsx
```diff
@@ -0,0 +1,285 @@
...
+  preview: string;
+}
+
+interface Props {
+  backgroundElement?: JSX.Element;
+  buttonTitleUpload?: string;
+  buttonTitleRemove?: string;
+  buttonTitleCancel?: string;
+
+  handleDrop: (files: File) => void;
```
fionawhim: `DropzoneFile`? Also, should either be called `file` and stay singular, or changed to an array.

modules-js/react-fleet/src/form-elements/UploadPhoto.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+  private onDrop = (files: File[]): void => {
+    this.setState(
+      {
+        file: Object.assign(files[0], {
+          preview: URL.createObjectURL(files[0]),
+        }),
+      },
+      () => {
+        this.state.file && this.props.handleDrop(this.state.file);
```
fionawhim, line 82: 👍

modules-js/react-fleet/src/form-elements/UploadPhoto.tsx
```diff
@@ -0,0 +1,285 @@
...
+    this.setState({ file: null }, () => {
+      this.props.handleRemove();
+    });
+  };
+
+  private handleButtonClick = (): void => {
+    const isUploading =
+      this.props.uploadProgress && this.props.uploadProgress < 100;
+
+    if (!this.dropzoneRef) {
```
fionawhim: `this.dropzoneRef` should always be truthy. Do you mean `.current`?

modules-js/react-fleet/src/form-elements/UploadPhoto.tsx
```diff
@@ -0,0 +1,285 @@
...
+  componentDidMount() {
+    this.getPreviewElementInformation();
+
+    window.addEventListener('resize', this.getPreviewElementInformation);
+  }
+
+  componentWillUnmount() {
+    // Make sure to revoke the data uris to avoid memory leaks
+    if (this.state.file) {
+      URL.revokeObjectURL(this.state.file.preview);
```
fionawhim, line 123: TIL about this function.

Do we want to do this, though? Aren't there times we'd want to use the preview outside of this component?

modules-js/react-fleet/src/form-elements/UploadPhoto.tsx
```diff
@@ -0,0 +1,285 @@
...
+                  ) : (
+                    <>
+                      {this.props.backgroundElement ||
+                        defaultInitialAppearance()}
+                    </>
+                  )}
+                </div>
+              </div>
+            )}
+          </Dropzone>
```
fionawhim, line 184: 👍

modules-js/react-fleet/src/form-elements/UploadPhoto.tsx
```diff
@@ -0,0 +1,285 @@
...
+    fill: 'none',
+    stroke: 'currentColor',
+  },
+});
+
+const PREVIEW_CONTAINER_STYLING = css({
+  position: 'relative',
+  cursor: 'pointer',
+  overflow: 'hidden',
+  backgroundColor: 'currentColor',
```
fionawhim, line 254: HOW HAVE I NEVER HEARD OF `currentColor` BEFORE??? Love it.

services-js/registry-certs/client/birth/components/SupportingDocumentsInput.tsx
```diff
@@ -0,0 +1,276 @@
...
+
+  // Check to see if a file does not exceed the file size limit, if defined.
+  private checkFileSize = (file: File): boolean | void => {
+    const sizeLimit = handleBytes.convert(this.props.sizeLimit);
+    const fileSize = handleBytes.format(file.size);
+
+    if (file.size > sizeLimit) {
+      // todo: consider handling via callback: https://github.com/CityOfBoston/digital/pull/41#discussion_r218545231
+
+      alert(
```
fionawhim, line 69: TODO: don't use `alert`

I also wonder about returning `false` from a file-upload callback to tell the picker to throw away the selection.

services-js/registry-certs/client/birth/components/SupportingDocumentsInput.tsx
```diff
@@ -0,0 +1,276 @@
...
+
+  private updateFiles = (files: FileList): void => {
+    const { selectedFiles } = this.state;
+
+    // Create an array from the FileList.
+    const fileArray: File[] = [];
+
+    for (let i = 0; i < files.length; i++) {
+      // Do not allow duplicate filenames, or files that are too large.
+      if (isUnique(files[i]) && this.checkFileSize(files[i])) {
```
fionawhim: This doesn't de-dupe file names that get added at the same time.

services-js/registry-certs/client/birth/components/SupportingDocumentsInput.tsx
```diff
@@ -0,0 +1,276 @@
...
+        // Clear actual input after each change.
+        if (this.inputRef.current) {
+          this.inputRef.current.value = null as any;
+        }
+      }
+    );
+
+    // Check to see if a file is already in this.state.selectedFiles.
+    function isUnique(newFile: File): boolean {
+      const result = selectedFiles.filter(file => file.name === newFile.name);
```
fionawhim: `find` is probably a better fit that `filter` for this algorithm, though the expected array sizes are low enough that there’s no practical difference.

services-js/registry-certs/client/birth/components/SupportingDocumentsInput.tsx
```diff
@@ -0,0 +1,276 @@
...
+
+      return result.length === 0;
+    }
+  };
+
+  // Clear a file from the list.
+  private clearFile(fileToClear: File): void {
+    const filteredList = this.state.selectedFiles.filter(file => {
```
fionawhim: You should be able to do straight equality of the `File` objects rather than rely on the file name.

services-js/registry-certs/client/birth/components/VerifyIdentificationComponent.tsx
```diff
@@ -1,20 +1,224 @@
...
+
+import FieldsetComponent from './FieldsetComponent';
+import SupportingDocumentsInput from './SupportingDocumentsInput';
+import IdIcon from '../icons/IdIcon';
+
+import { SECTION_HEADING_STYLING, SUPPORTING_TEXT_STYLING } from '../styling';
+
+// todo: props to pass addt’l info in if Registry has initiated this
+interface Props {
+  supportingDocuments: (documents: File[]) => void;
```
fionawhim: These names could be clearer. Is this what gets called back when they’re added?

services-js/registry-certs/client/birth/components/VerifyIdentificationComponent.tsx
```diff
@@ -1,20 +1,224 @@
...
+  handleBooleanChange = (event: React.ChangeEvent<HTMLInputElement>): void => {
+    this.setState({
+      requireSupportingDocuments: event.target.value === 'yes',
+    });
+  };
+
+  // Inform the parent whether or not all required information has been added
+  // by the user, if necessary.
+  componentDidUpdate(_prevProps: Readonly<Props>, prevState: Readonly<State>) {
+    if (this.props.isComplete && prevState !== this.state) {
```
fionawhim: the `this.props.isComplete` check here feels like it’s not `componentDidUpdate`’s concern. That can be handled in `checkIsComplete`.

Is it ok that `prevState !== this.state` isn’t a deep equals?

modules-js/react-fleet/package.json
jessicamarcus: Ah, we were talking about polyfills and I got distracted/confused!

modules-js/react-fleet/src/form-elements/UploadPhoto.tsx
jessicamarcus: thx for the catch!

modules-js/react-fleet/src/form-elements/UploadPhoto.tsx
jessicamarcus: I was going by their example: https://react-dropzone.netlify.com/#!/Previews - I should have said so, and included a link back to their docs in a comment.

modules-js/react-fleet/src/form-elements/UploadPhoto.tsx
jessicamarcus: It is the best thing ever. The **best**

services-js/registry-certs/client/birth/components/SupportingDocumentsInput.tsx
jessicamarcus: Ah right, the polyfill changes you made should pick that up automatically...

services-js/registry-certs/client/birth/components/SupportingDocumentsInput.tsx
jessicamarcus: There really isn't a way to add two files with the same name at one time with the system file picker dialog.

services-js/registry-certs/client/birth/components/VerifyIdentificationComponent.tsx
jessicamarcus: Ah, okay - I figured if this component was being used by a parent that didn’t care about `isComplete`, then `checkIsComplete` will never be called... but I guess it doesn’t matter **that** much!

`prevState !== this.state` is fine in this case because we don’t care _what_ is in the state, only that it has changed in order to call `checkIsComplete`.

services-js/registry-certs/client/birth/components/SupportingDocumentsInput.tsx
jessicamarcus: There are a couple of todos regarding this; it’s definitely something I’d like to come back to later.

fionawhim (approved):

PR #314, opened 2019-02-14
Fixes for birth certificate receipt

fionawhim:
- Updates uncaptured fixture to be an event, not a charge, to match the
   successful fixture
 - Works around null subtotal and total fields for birth certs
 - Fixes units for costs


jessicamarcus (approved):
👍

PR #315, opened 2019-02-14
Adjustments to header appearance/behavior

### LINKED ISSUE #311: Adjust React header component behavior
Header behavior on scroll should be in line with that of boston.gov. On large viewports, the City seal appears in the header element. When a user scrolls more than ~310px and then stops, the seal moves up in the header so it is visually half-hidden. If the user scrolls back to the top of the page, the seal returns to its original position. Transition is accomplished by adding/removing the `s--u` css class.

The projects in the `digital` repo are built with a header component that uses an automatically-extracted version of the boston.gov header, and it does not have this behavior. Also, when an app does not have a secondary nav component, the seal’s initial position ends up hanging over the breadcrumbs oddly.

jessicamarcus:
City seal show/hide behavior on scroll should be in line with that of boston.gov, and additional vertical space should be added if secondaryNav is not present.

Closes #311


modules-js/react-fleet/src/components/Header.tsx
```diff
@@ -6,10 +6,47 @@ export default function Header(
...
+      } else {
+        sealElement.classList.remove('s--u');
+      }
+    } else {
+      // The listener is unnecessary if the seal is not visible, so remove it.
+      window.removeEventListener('scroll', handleScroll);
+    }
+  }
+
+  window.addEventListener('scroll', handleScroll);
```
fionawhim: This is likely adding a bunch of event listeners over the course of the app.

I think we need to do this in a hook or change to a class component and do it with the lifecycle events.

modules-js/react-fleet/src/components/Header.tsx
```diff
@@ -22,3 +59,12 @@ export default function Header(
     </>
   );
 }
+
+function isSealVisible(): boolean {
```
fionawhim, line 73: Make it clear that the visibility is more about breakpoints than position on the screen.

modules-js/react-fleet/src/layouts/AppLayout.tsx
```diff
@@ -45,7 +45,11 @@ export default function AppLayout({
 
           {secondaryNav}
 
-          {breadcrumbNav}
+          {secondaryNav ? (
+            <>{breadcrumbNav}</>
+          ) : (
+            <div className="p-t300">{breadcrumbNav}</div>
```
fionawhim, line 51: I think this is unnecessary now?

fionawhim (changes requested):

modules-js/react-fleet/src/layouts/AppLayout.tsx
jessicamarcus: My understanding was that the Seal sometimes will overlap with the breadcrumb text when the secondary nav isn’t present, which is a separate concern from the scrolling behavior.

fionawhim (approved):

fionawhim:
🎣 !!!

modules-js/react-fleet/src/components/Header.tsx
```diff
@@ -21,22 +22,19 @@ export default function Header(
...
-
-          handleScrollStop();
-
-          isScrolling = false;
-        });
-
-        isScrolling = true;
+      if (timer) {
+        // Update scroll position while user is scrolling
+        window.requestAnimationFrame(() => setScrollPosition(lastPos));
```
fionawhim: I think you don't need to keep `scrollPosition` around anymore, right? You can just check `window.scrollY` inside of `handleScrollStop`. Should prevent some re-renderings if you're not changing state.

jessicamarcus: hah, you’re right - I was being myopic.

fionawhim (approved):

PR #318, opened 2019-02-15
Adds an "order shipped" email for birth certs

fionawhim:
- Moves birth / death specific logic out of the handlebars templates
   and instead passes in values set up in EmailTemplates.ts
 - Corrects the database date response type and moves date formatting to
   where it's appearing
 - Makes emails "from" either "registry@boston.gov" or
   "birth@boston.gov"


jessicamarcus (approved):
lgtm! 🎉

PR #324, opened 2019-02-20
Upgrade to Jest 24.1

fionawhim:
This takes advantage of Jest 24’s better support of using TypeScript
through Babel. Updates all packages to run TypeScript through Babel,
rather than through custom loaders. This means putting .babelrc files
everywhere.

This is the first step towards doing all compilation and running with
Babel, leaving TypeScript’s tsc just for type checking.

Removes config-jest-babel and config-jest-typescript because Jest seems
unable to find .babelrc files when using config presets. (And Jest 24
makes most of the config unnecessary.)

Updates some typing around Jest Mocks to match the stricter types
in Jest 24. Seen especially in stripe-events.test.ts

Also:
 - Pulls in new templates for 311 and updates the snapshots
 - Changes next-client-common to be a dual ESM / ES5 package like
   react-fleet and form-common
 - Rewrites many of the READMEs around our compilation and library
   strategies
 - Attempts — without success — to remove a peer dependency warning by
   sprinkling core-js dependencies everywhere.
 - Removes the separate object spread operator Babel transform, since
   that is now the default in preset-env.
 - importing .snap files now evaluates them, so we don’t need to create
   the Function and execute it

registry-certs staging seems to have lost between 1–4k per JS file due
to these changes.


fionawhim:
sorry

README.md
```diff
@@ -127,8 +127,8 @@ something gets added that _does_ demand testing.
 
 #### Server-side Libraries
 
-For server-side code, we use TypeScript directly. That means that the build
-steps are handled by `tsc` and development uses `tsc-watch`. Projects’
+For server-side code, we build with TypeScript directly. That means that the
```
jessicamarcus, line 130: docs! 👍

jessicamarcus (approved):
💯 Thanks for taking the care to update documentation! Lots of awesome tidying and cleanup in this commit, which I’m super excited about.

PR #333, opened 2019-02-22
Adds file upload, delete functionality; resolves #265

### LINKED ISSUE #265: ID verification - add/remove files in DB
Binary files provided by the user will be immediately uploaded to the db server, separate from the other information collected during the questions workflow.

Functionality to delete a file on the server must be present for cases when the user wants to replace or delete a file before proceeding with their request order.

The server will return a unique `attachmentKey` on success.

_The following is no longer applicable:_

> This page supports the case of a Registry-initiated request for id verification to support an existing order. The user will land on this page and see the order number and name on the requested record, and the functional component. 
> 
> Page will handle uploading files to the db, and showing errors and/or success confirmation to the user, while the upload/photo functionality is handled by `VerifyIdentificationComponent`.
> 
> **Questions:**
> - should we display name of person who ordered in addition to the name on requested record and order id#?
> - need text content for submission success
> - should we get already-submitted images/files from server and display those assets as already selected in the “upload front/back of id” and “supporting documents” inputs?
> 
> **Functional requirements:**
> - submit files to db
> - show error messaging
> - show success messaging

jessicamarcus:
- creates `UploadPhoto` component in react-fleet
- freestanding “verify information” page is no longer a project requirement; deleted along with related Storybook stories
- moved some types around
- updated `SupportingDocumentsInput` component to reflect upload state of individual files
- uploading files to the server is now handled individually on a per-file basis: changed to accept a single item rather than an array
- added GraphQL query for file deletion
- added `nock` package for testing purposes


modules-js/react-fleet/src/form-elements/UploadPhoto.stories.tsx
```diff
@@ -40,6 +40,23 @@ storiesOf('Form Elements/Upload photo', module)
         handleRemove={() => {}}
       />
     </div>
+  ))
+  .add('displaying image File from data store', () => (
```
fionawhim, line 44: This label doesn't match with what’s going on below. Shouldn’t this story be showing `initialFile` and not the `errorMessage` case?

services-js/registry-certs/client/birth/components/SupportingDocumentsInput.tsx
```diff
@@ -1,22 +1,31 @@
...
+import { observer } from 'mobx-react';
+
 interface Props {
-  name: string;
-  title: string;
-  fileTypes?: string[] | string;
-  sizeLimit: FileSize;
-  handleChange(documents: File[]): void;
+  uploadSessionId: string;
+  selectedFiles?: UploadableFile[];
```
fionawhim: If you require this to always exist, it would allow you to remove some existence checks down below. (I would assume callers would pass it an empty array on first load.)

services-js/registry-certs/client/birth/components/SupportingDocumentsInput.tsx
```diff
@@ -85,65 +82,79 @@ ${file.name} is ${fileSize.amount.toFixed(2) +
...
-      this.props.handleChange(this.state.selectedFiles);
-    });
-  }
+    if (selectedFiles) {
+      const file = selectedFiles.find(
+        fileObject => fileObject.file === fileToClear
+      );
+
+      if (file) {
+        file.delete(didCancel).then(() => {
```
fionawhim: Prefer `async`/`await` unless there’s a specific reason to go with `then`. As you saw when we worked on the `delete` method, it is easy to get mixed up over the order that the `then` block executes relative to any statements that come after it.

services-js/registry-certs/client/birth/components/SupportingDocumentsInput.tsx
```diff
@@ -85,65 +82,79 @@ ${file.name} is ${fileSize.amount.toFixed(2) +
...
+          );
+        });
+      }
+    }
+  };
 
   private handleFileChange = (): void => {
     if (this.inputRef.current && this.inputRef.current.files) {
-      this.updateFiles(this.inputRef.current.files);
+      this.addFilesFromInput(this.inputRef.current.files);
```
fionawhim, line 142: I like this name change.

services-js/registry-certs/client/birth/components/SupportingDocumentsInput.tsx
```diff
@@ -153,45 +164,116 @@ ${file.name} is ${fileSize.amount.toFixed(2) +
...
     );
   }
 }
 
+interface FileButtonProps {
+  uploadableFile: UploadableFile;
+  deleteFile: (file: File, didCancel?: boolean) => void;
+}
+
+function FileButton(props: FileButtonProps): JSX.Element {
```
fionawhim, line 214: Wrap this in `observer` so it sees changes to progress and such

services-js/registry-certs/client/birth/components/VerifyIdentificationComponent.tsx
```diff
@@ -13,11 +14,17 @@ import FieldsetComponent from '../components/FieldsetComponent';
...
+
 import { SECTION_HEADING_STYLING, SUPPORTING_TEXT_STYLING } from '../styling';
 
 interface Props {
+  requestInformation: BirthCertificateRequestInformation;
   sectionsToDisplay?: 'all' | 'supportingDocumentsOnly';
-  updateSupportingDocuments: (documents: File[]) => void;
+  uploadSessionId: string;
+  supportingDocuments?: UploadableFile[];
+  updateSupportingDocuments: (documents: UploadableFile[]) => void;
```
fionawhim, line 27: I would expect that `uploadSessionId` and `supportingDocuments` would come off of the `requestInformation` prop. I assume this is a vestige of when this component would have to work w/o a `BirthCertificateRequestInformation` object?

I think it should be cleaned up, but it can be done later.

services-js/registry-certs/client/models/UploadableFile.test.ts
```diff
@@ -0,0 +1,100 @@
...
+
+jest.mock('../queries/delete-birth-certificate-uploaded-file');
+
+import deleteBirthCertificateUploadedFile from '../queries/delete-birth-certificate-uploaded-file';
+const deleteBirthCertificateUploadedFileMock: jest.MockInstance<
+  ReturnType<typeof deleteBirthCertificateUploadedFile>,
+  Parameters<typeof deleteBirthCertificateUploadedFile>
+> = deleteBirthCertificateUploadedFile as any;
+
+const serverSuccessResponse = '{"filename":"sample.pdf","attachmentKey":"17"}';
```
fionawhim: It would be good to have this defined as an object literal that you can put the `UploadResponse` type on.

services-js/registry-certs/client/models/UploadableFile.ts
```diff
@@ -0,0 +1,190 @@
+import 'isomorphic-fetch';
```
fionawhim: You can assume this is already provided.

services-js/registry-certs/client/models/UploadableFile.ts
```diff
@@ -0,0 +1,190 @@
...
+
+    if (!(xhr instanceof XMLHttpRequest)) {
+      return;
+    }
+
+    try {
+      json = JSON.parse(xhr.responseText);
+    } catch (e) {
+      this.status = 'uploadError';
+      this.errorMessage = xhr.statusText;
```
fionawhim, line 130: should probably `return` here since JSON not parsing is an error in and of itself.

services-js/registry-certs/client/models/UploadableFile.ts
```diff
@@ -0,0 +1,190 @@
...
+
+    uploadRequest.onload = this.handleLoad;
+    uploadRequest.onerror = this.handleError;
+    uploadRequest.onloadend = this.handleLoadEnd;
+
+    if (uploadRequest.upload) {
+      uploadRequest.upload.onprogress = this.handleProgress;
+    }
+
+    this.progress = 0;
```
fionawhim, line 75: Good. Important to reset this for the retrying case.

fionawhim (approved):

fionawhim:
There are a few things to tweak. Get as much as you can done and then merge so I can work off of this on Monday.

modules-js/react-fleet/src/form-elements/UploadPhoto.stories.tsx
jessicamarcus: This test is incomplete, because I was trying to use an actual photograph and was going to try doing so by pulling the `File` from the DOM, and then inserting it... would love to talk about this!

PR #337, opened 2019-02-26
Fixes Access Boston logout

fionawhim:
Changes to use SLO.saml2 for path to the logout response endpoint.
(startSLO.ping is for browsers to visit directly)


jessicamarcus (approved):
👍

PR #338, opened 2019-02-26
Fixes answers not getting saved

fionawhim:
Adds MouseEvent argument to all of the event handlers since they will be
passed an event as their first argument.

QuestionsPage was assuming that they wouldn’t be passed an argument, so
advanceQuestion had an optional first argument instead.

### DIFF
--- a/services-js/registry-certs/client/birth/questions/ParentalInformation.tsx
+++ b/services-js/registry-certs/client/birth/questions/ParentalInformation.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { MouseEvent } from 'react';
 import { observer } from 'mobx-react';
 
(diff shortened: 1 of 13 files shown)


services-js/registry-certs/client/birth/questions/ParentalInformation.tsx
```diff
@@ -208,8 +210,8 @@ export default class ParentalInformation extends React.Component<Props> {
             <p>
               <strong>
                 If you are listed on the record, you will need to provide a
-                valid form of identification (i.e. driver’s license, state ID,
-                military ID, or passport) to get a copy.
+                valid form of identification (<i>e.g.</i> driver’s license,
```
jessicamarcus, line 213: This fix makes me happy.

jessicamarcus (approved):
👍

PR #340, opened 2019-02-27
Fixes for preview image when removing

fionawhim:
- Revokes the URL to free up memory
 - Sets state to null so that it stops showing


jessicamarcus (approved):
lgtm!

PR #343, opened 2019-02-27
Birth certs polish items

fionawhim:
- Limits uploadable file types to images we're likely to get from
   scanners
 - Adds spacing around the date fields on mobile
 - Limits font size of the heading so that we don’t wrap
 - Tweaks the "forSelf" icon so that it’s consistent with "for someone
   else" when they collapse into a column on mobile


jessicamarcus (approved):
👍

PR #344, opened 2019-02-28
Sends AddPayment to DB for birth cert changes

fionawhim:
Lets the fulfillment database keep track of the Stripe charge ID for
these transactions.


jessicamarcus (approved):
lgtm!

PR #346, opened 2019-03-05
MFA device changes

fionawhim:
- Adds banner when the user will need to register an MFA device in the
   future but it’s not required right now
 - Only shows the "manage MFA devices" link if the user has a device

Uses a new assertion attribute, isUserRegistered, to tell if the user
has an MFA device already. If they do not, allows them to complete the
MFA registration process even if the "needsMfa" flag hasn’t yet been
set.

This is necessary to support a gradual roll-out of the MFA registration
requirements.

See CityOfBoston/iam#428 and CityOfBoston/iam#411


services-js/access-boston/src/pages/index.tsx
```diff
@@ -33,6 +35,8 @@ interface Props {
   account: Account;
   apps: Apps;
   flashMessage?: FlashMessage;
+  /** Used for calculating relative dates so that snapshots don’t change with Date.now() */
```
jessicamarcus, line 38: 💯

jessicamarcus (approved):
lgtm!

PR #347, opened 2019-03-05
Changes to put birth request details in item

fionawhim:
Per Rich / Scott, we want to attach these notes details to the birth
certificate item table, not as notes on the order.


jessicamarcus (approved):
👍

PR #350, opened 2019-03-06
Runs prepare as part of pretest

fionawhim:
Necessary to bundle install before running tests.


fionawhim:
Self-approved to fix build.

PR #354, opened 2019-03-08
Fix for some dropped contact emails

fionawhim:
- Stops sending the message as an email header, since that’s redundant
   and not all messages can parse as headers
 - Updates rollbar gem to include the fix that catches exceptions in
   ApplicationMailer subclasses


mmcgowanBOS (approved):
Nice.

PR #355, opened 2019-03-11
Rewrites "unsubscribe" subject lines

fionawhim:
Prevents Google Groups from eating the message rather than sending it
along to the distribution list.


mmcgowanBOS (approved):
lgtm.

PR #357, opened 2019-03-12
Allows visiting /mfa if account has no MFA

fionawhim:
Switches logic from "needs" MFA (which is that the requirement is set)
to "not having MFA" so that the "complete it now" banner will work.

See CityOfBoston/iam#498


jessicamarcus (approved):
👍

PR #359, opened 2019-03-12
Adds a 10 minute duration for forgot password

fionawhim:
Matches the behavior where the auth token expires after 10 minutes.

Changes from Boom.forbidden to error codes so that the frontend can
interpret what’s going on.


jessicamarcus (approved):
lgtm 😄

PR #362, opened 2019-03-13
Upgrades Rails apps to 5.1.6.2 for latest patches

fionawhim:
Change to config.i18n.fallbacks per a message during upgrade.


mmcgowanBOS (approved):
lgtm.

Issue #352, opened 2019-03-07
Add Google Analytics tags


rachelbraun:
Track:

1. Who they're order a birth certificate for (including all someone else options)
2. Were they born in Boston
3. Did the parents live in Boston at the time of the birth
4. What year were they born
5. Were parents married at time of the birth
6. Clicked upload front of ID
7. Clicked upload supporting documents
8. Clicked 'No ID?' link
9. Qty of birth records requested
10. Shipping address city (text box) and state (dropdown)

rachelbraun:
Adding an 11th.

Track:
- [ ] Who they order a birth certificate for (including all someone else options)
- [ ] Were they born in Boston
- [ ] Did the parents live in Boston at the time of the birth
- [ ] What year were they born
- [ ] Were parents married at the time of the birth
- [ ] Clicked upload front of ID
- [ ] Clicked upload supporting documents
- [ ] Clicked 'No ID?' link
- [ ] Qty of birth records requested
- [ ] Shipping address city (text box) and state (dropdown)
- [ ] How many users navigated to the Purchasing for a Client page

rachelbraun:
Adding one post-sync.

Track:
 

- [x] Who they order a birth certificate for (including all someone else options)
- [x] How many users navigated to the Purchasing for a Client page
- [x] Were they born in Boston
- [x] Did the parents live in Boston at the time of the birth
- [x] What year were they born
- [x] Were parents married at the time of the birth
- [x] Clicked upload front of ID
- [x] Clicked upload supporting documents
- [x] Clicked 'No ID?' link
- [x] Qty of birth records requested
- [x] Shipping address city (text box) and state (dropdown)

PR #364, opened 2019-03-13
Adds many of the required GA items

jessicamarcus:
See https://github.com/CityOfBoston/digital/issues/352

Although my browser debugger plugin is showing analytics commands being sent, I don’t see any Birth Certificate products in the ecommerce section of GA? I do see Events coming in, though.

Because of the launch, I wanted to get at least some/most of the analytics bits in as soon as possible...

### DIFF
--- a/services-js/registry-certs/client/birth/QuestionsPage.tsx
+++ b/services-js/registry-certs/client/birth/QuestionsPage.tsx
@@ -28,7 +28,7 @@ interface InitialProps {
 
 interface Props
   extends InitialProps,
-    Pick<PageDependencies, 'birthCertificateRequest'> {}
+    Pick<PageDependencies, 'birthCertificateRequest' | 'siteAnalytics'> {}
 
 interface State {
   /**
@@ -87,7 +87,9 @@ export default class QuestionsPage extends React.Component<Props, State> {
     if (props.currentStep !== state.currentStep) {
       return {
         currentStep: props.currentStep,
-        localBirthCertificateRequest: props.birthCertificateRequest.clone(),
+        localBirthCertificateRequest: props.birthCertificateRequest.clone(
+          props.siteAnalytics
+        ),
       };
     } else {
       return null;
@@ -98,7 +100,9 @@ export default class QuestionsPage extends React.Component<Props, State> {
     super(props);
 
     this.state = {
-      localBirthCertificateRequest: props.birthCertificateRequest.clone(),
+      localBirthCertificateRequest: props.birthCertificateRequest.clone(
+        props.siteAnalytics
+      ),
       currentStep: props.currentStep,
     };
   }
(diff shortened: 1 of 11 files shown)


services-js/registry-certs/client/birth/QuestionsPage.tsx
```diff
@@ -147,15 +161,60 @@ export default class QuestionsPage extends React.Component<Props, State> {
...
+      if (bornInBoston !== 'yes') {
+        result += `; parents lived in Boston: ${parentsLivedInBoston}`;
+      }
+
+      return result;
+    }
+  };
+
+  gaSendEvent = (label: string): void => {
+    this.props.siteAnalytics.sendEvent('click', {
```
fionawhim, line 199: I'd recommend changing this so that you have separate actions for each kind of question, and then use the label as the answer to that question.

That way you can filter down per question.

services-js/registry-certs/client/birth/ReviewRequestPage.tsx
```diff
@@ -47,12 +63,23 @@ export default class ReviewRequestPage extends React.Component<Props> {
...
 
+    siteAnalytics.addProduct(
+      '0',
+      'Birth certificate',
+      'Birth certificate',
+      this.props.birthCertificateRequest.quantity,
+      BIRTH_CERTIFICATE_COST / 100
+    );
+
+    siteAnalytics.setProductAction('remove');
```
fionawhim, line 81: 👍

services-js/registry-certs/client/birth/components/VerifyIdentificationComponent.tsx
```diff
@@ -107,7 +112,10 @@ export default class VerifyIdentificationComponent extends React.Component<
         </ul>
 
         <div className="g">
-          <div className="g--6 m-v500">
+          <div
+            className="g--6 m-v500"
+            onClick={() => this.handleSendEventClick('upload front of ID')}
```
fionawhim, line 117: just double-checking that these are all running (that the event’s bubbling wasn't stopped).

If so, 👍

services-js/registry-certs/client/birth/QuestionsPage.tsx
jessicamarcus: Awesome, this syncs with where my head is at.

fionawhim:
OK I solved the GA issue. The problem is that `clone` was being called during component mount, while `attach` only happened _after_ in `componentDidMount`. So the first clone never got the `siteAnalytics`.

I'd recommend changing to set `siteAnalytics` right after creation in `_app.tsx`. And maybe rename the method from `attach` to `setSiteAnalytics` to avoid confusion with the attachment pattern.

fionawhim (approved):

PR #366, opened 2019-03-14
Enforces "+1" for MFA phone numbers

fionawhim:
Gets around a bug where Ping can't send to 857 area codes (a common
Boston cell number) without a country code.

See CityOfBoston/iam#504


jessicamarcus (approved):
lgtm!

PR #367, opened 2019-03-14
Configure Ruby apps not to send 404s to Rollbar

fionawhim:
They’re junk exceptions that we don’t care about.


mmcgowanBOS (approved):
Nice.

PR #369, opened 2019-03-14
Removes timezone from birthdate display in email text

### LINKED ISSUE #368: Fix (timezone) birthdate issue
Date block is changing some birth dates to a day earlier due to timezone discrepancies.

Case 1:
User types in April 4 1970
Request shows as April 3 1970
Yarmouth, ME
Submitted at 10:40am on 3/8

Case 2:
User types in Oct 14 1947
Request shows as Oct 13 1947
Newtonville, MA
Submitted at 8:40am on 3/14

jessicamarcus:
Fixes #368 

Registry confirmed that this bug was only occurring in the email text to the customer; I removed the `.tz()` from the momentjs formatting, since a timezone isn’t relevant in the case of a birth date.

### DIFF
--- a/services-js/registry-certs/server/email/EmailTemplates.ts
+++ b/services-js/registry-certs/server/email/EmailTemplates.ts
@@ -146,9 +146,9 @@ export class EmailTemplates {
         items: receipt.items.map(({ cost, quantity, name, date }) => ({
           quantity,
           cost,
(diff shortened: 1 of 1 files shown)


fionawhim (approved):

fionawhim:
Let's test on staging before shipping tho

PR #376, opened 2019-03-15
Enhances GA reporting; improves appearance of control focus states

jessicamarcus:
- adds clear focus states to the dropzone input container as well as the upload photo/remove photo buttons
- adds more granularity to GA events reporting (differentiates between button clicks and uploads via drag+drop)

Although I am currently working on broader GA tasks, I thought it would be tidier in the end if this group of changes were in its own PR.


fionawhim (approved):
👍

PR #377, opened 2019-03-15
Switches to use UTC exclusively for date input

### LINKED ISSUE #374: Birth Certs broken on iOS 9
Safari 9 seems not to support `Intl`

https://caniuse.com/#search=Intl

https://rollbar.com/CityofBoston/registry-certs/items/143/

### LINKED ISSUE #375: Safari: Date wrong on review page
Report from a constituent that the date on the review page is one day off from their actual date.

fionawhim:
We were running into inconsistencies in how Safari created dates in the
past that were a different daylight savings from the current time.

Removes string inputs for MemorableDateInput for now. If we want them in
the future we can put them back and figure more out about how they
should work with timezones.

Also removes uses of Intl, which isn’t present in iOS 9.

Fixes #375
Fixes #374


jessicamarcus (approved):
👍 I can’t wait for the world to stop using time zones entirely.



(assuming we don’t destroy it first)

fionawhim:
We'd need to time travel back and destroy the timezones of the past, too.

PR #379, opened 2019-03-15
Removes quotes from example for looking up email

fionawhim:
At least this morning, the quotes became part of the string in the task,
which then didn't match any addresses.


mmcgowanBOS (approved):
Looks good to me.

PR #380, opened 2019-03-18
Aggressive no-cache headers for Access Boston

fionawhim:
Keeps Firefox back/forward cache from letting you "back" from the
logged-out landing page into a still-logged in web portal page.

May also keep other browsers from being so cache-y.

See CityOfBoston/iam#502

Also fixes logout button to be "submit" and fixes JSX httpEquiv
attribute casing in _document.

### DIFF
--- a/services-js/access-boston/src/server/access-boston.ts
+++ b/services-js/access-boston/src/server/access-boston.ts
@@ -4,7 +4,7 @@ import fs from 'fs';
 import path from 'path';
 import { promisify } from 'util';
 
-import { Server as HapiServer } from 'hapi';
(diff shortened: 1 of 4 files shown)


services-js/access-boston/src/server/access-boston.ts
```diff
@@ -392,6 +392,18 @@ async function addNext(server: HapiServer) {
     return h.continue;
   };
 
+  // Hapi’s default caching handling doesn’t support things like "no-store,"
```
jessicamarcus, line 395: Good documentation 👍

jessicamarcus (approved):
lgtm

PR #383, opened 2019-03-18
Switches from includes to indexOf check

fionawhim:
IE11 doesn’t have includes, so we want to play it safe and not rely on
any other polyfilling.

### DIFF
--- a/modules-js/next-client-common/polyfills.js
+++ b/modules-js/next-client-common/polyfills.js
@@ -7,17 +7,26 @@
  * "usage" should cause Babel to automatically include necessary polyfills.
  */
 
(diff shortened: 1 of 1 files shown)


modules-js/next-client-common/polyfills.js
```diff
@@ -7,17 +7,26 @@
  * "usage" should cause Babel to automatically include necessary polyfills.
  */
 
-// Issues arise in IE11: Stencil’s polyfill and core-js’ polyfill compete with each other in a neverending loop.
+// Issues arise in IE11: Stencil’s polyfill and core-js’ polyfill compete with
```
jessicamarcus, line 10: Thanks for expanding on this!

jessicamarcus (approved):

Issue #339, opened 2019-02-26
QA Fix: Cancel and Start Over Lost Path


rachelbraun:
Identified a lost path:

If I fill out my info, then hit 'cancel and start over' on the 'review record request' page, I get taken back to the start (with information cleared as it should).

If at this point, I choose to hit the 'back' on my browser, I'm taken back to the 'review record request' page but with no birth certificate information specified. 

I can continue forward to make a purchase until I hit 'submit order' - at which point I get an error pop up.

rachelbraun:
![birth certs review your record request page](https://user-images.githubusercontent.com/35377030/53447417-14277380-39e3-11e9-8c16-f05386806ea9.png)

<img width="855" alt="screen shot 2019-02-26 at 4 23 37 pm" src="https://user-images.githubusercontent.com/35377030/53447491-35885f80-39e3-11e9-8cc8-d3ced7c6f292.png">

<img width="855" alt="screen shot 2019-02-26 at 4 23 47 pm" src="https://user-images.githubusercontent.com/35377030/53447494-37eab980-39e3-11e9-88e1-68d92b585cf5.png">

<img width="855" alt="screen shot 2019-02-26 at 4 23 56 pm" src="https://user-images.githubusercontent.com/35377030/53447499-3ae5aa00-39e3-11e9-83b6-996c071ea728.png">

<img width="854" alt="screen shot 2019-02-26 at 4 24 08 pm" src="https://user-images.githubusercontent.com/35377030/53447503-3d480400-39e3-11e9-9614-0c84e73fd866.png">

PR #384, opened 2019-03-19
Redirects user to /birth if page reload occurs

jessicamarcus:
Quick fix to handle cases where a user is on `/review` or `/checkout`, but the question flow results have been lost to a page reload.


fionawhim (approved):

fionawhim:
Awesome! Thanks!

jessicamarcus:
Resolves #339.

PR #385, opened 2019-03-19
Prevents onDrop setState() if user drag-and-drops incorrect filetype

jessicamarcus:
Resolves https://rollbar.com/CityofBoston/registry-certs/items/140/

The error message `Failed to execute 'createObjectURL' on 'URL'` occurs after the user tries to drag-and-drop a non-image file into the dropzone; preventing `setState()` from being called in those cases prevents the error.

Issue is discussed in this thread: https://github.com/react-dropzone/react-dropzone/issues/255


fionawhim (approved):
Nice find.

PR #387, opened 2019-03-19
Moves route change into setState callback

jessicamarcus:
Resolves https://rollbar.com/CityofBoston/registry-certs/items/154/

When an order has been submitted, we need to ensure that state updates have been completed before the route changes.

### DIFF
--- a/services-js/registry-certs/client/birth/CheckoutPage.tsx
+++ b/services-js/registry-certs/client/birth/CheckoutPage.tsx
@@ -240,20 +240,23 @@ export default class BirthCheckoutPage extends React.Component<Props, State> {
     birthCertificateRequest.clearBirthCertificateRequest();
     orderProvider.clear();
 
-    this.setState({
-      order: await orderProvider.get(),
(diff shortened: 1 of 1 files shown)


fionawhim:
What does the `CheckoutPage` end up looking like in this case? It seems that `render` might flash a blank order of some kind.

jessicamarcus:
I didn’t notice anything jarring, but I could push it to staging and see what it looks like to Bruti (and whomever)?

fionawhim:
One good test is to remove the `Router.push` call and see what happens to the UI.

Then you can decide if you want to keep that, or remove the `setState` entirely.

jessicamarcus:
Ah, good call. When I tried removing the `setState`, the UI would hang. 🤔

jessicamarcus:
I’ve done some more testing and the transition from “order submitting” to the confirmation page is visually instantaneous. I tried a throttled connection, same result.

fionawhim (approved):

PR #388, opened 2019-03-20
Adds source map uploading of Next to Rollbar

fionawhim:
Also removes "guess_uncaught_frames" because we don’t minimize with
newlines.


jessicamarcus (approved):
❤️ sourcemaps! ❤️

PR #390, opened 2019-03-21
Adds s3 deploying to Next apps

fionawhim:
Updates hapi-next to support external asset prefixes. Also adds a
"download" mode to sending source maps to Rollbar, to accommodate when
the source maps are too large for the upload API.


jessicamarcus (approved):
lgtm!

PR #393, opened 2019-03-22
Keeps post-submit errors from being swallowed

fionawhim:
Ensures that we’re seeing everything that might be going wrong during
the submission process.

Also increases the range for order IDs to reduce collisions.


fionawhim:
Self-approvin' since this catches production things.

PR #401, opened 2019-03-27
Debugging and logging changes around SF disconnect

fionawhim:
- Changes subscriptions to happen for all successful handshakes, rather
   than just the first
 - Uses "/meta/disconnect" messages as the signal to restart the indexer
 - Adds debug logging and sends messages to Rollbar for more insight
   into what might be causing this
 - Changes "metadata" to "extra" for Rollbar because "metadata" may have
   its own meaning.


jessicamarcus (approved):
👍

PR #404, opened 2019-03-28
Adds protection around malformed MFA dates

fionawhim:
Also adds protection for the case where there’s an auth part of the
session but not session data (such as can happen when creating the
LoginSession fails).

Fixes https://rollbar.com/CityofBoston/access-boston/items/29/
Fixes https://rollbar.com/CityofBoston/access-boston/items/30/


jessicamarcus (approved):
lgtm!

PR #410, opened 2019-04-01
Removes the crumb protection from /logout

fionawhim:
We’re seeing errors go by and want to figure out what’s up before we
enforce this (slightly unnecessary) protection.

See: https://rollbar.com/CityofBoston/access-boston/items/49/


fionawhim:
Self-approve for prod fix

PR #413, opened 2019-04-01
Adds Boom exceptions’ "data" to Rollbar in GraphQL

fionawhim:
Previously the Boom data was only sent on 500 errors, not those caught
by GraphQL.


jessicamarcus (approved):
lgtm!

PR #415, opened 2019-04-01
Warning message for recent birthdates

### LINKED ISSUE #403: Add warning for parents ordering for newborns
If a user is looking for a birth cert for someone born in the last 3 weeks, pop up the same red module used for folks whose parents weren't married.

Text should say:

**We recommend waiting until the child is at least two weeks old before you request a birth certificate.** It generally takes two weeks for us to receive paperwork from the hospital.

If you order too early, we can only hold your request for seven days while we wait for the hospital record. If we don’t receive the record in time, your request will be canceled automatically and your card will not be charged. You’ll have to resubmit your request.

jessicamarcus:
Resolves #403 - small header added, and text adjusted based on conversation with James.

### DIFF
--- a/services-js/registry-certs/client/birth/QuestionsPage.stories.tsx
+++ b/services-js/registry-certs/client/birth/QuestionsPage.stories.tsx
@@ -78,9 +78,8 @@ storiesOf('Birth/QuestionsFlow/2. Born in Boston?', module)
     />
   ));
 
-storiesOf('Birth/QuestionsFlow/3. Personal information', module).add(
-  'blank',
-  () => (
+storiesOf('Birth/QuestionsFlow/3. Personal information', module)
+  .add('blank', () => (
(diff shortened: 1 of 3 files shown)


services-js/registry-certs/client/birth/QuestionsPage.stories.tsx
```diff
@@ -78,18 +78,28 @@ storiesOf('Birth/QuestionsFlow/2. Born in Boston?', module)
...
       siteAnalytics={{} as any}
       currentStep="personalInformation"
       birthCertificateRequest={makeBirthCertificateRequest({
         forSelf: true,
       })}
     />
-  )
-);
+  ))
+  .add('birth too recent', () => (
```
fionawhim, line 91: 👍

fionawhim (approved):
Awesome!

PR #417, opened 2019-04-02
Adds test prop; reorganizes Storybook stories for Birth

jessicamarcus:
Question components were documented as variations on the `QuestionsFlow` component, making it difficult to pass in test properties. I added and reorganized in order for each question to have its own Story, add missing Stories for a utility component, and to tidy up some.


fionawhim (approved):
🥇

PR #418, opened 2019-04-03
Modifications to speed up full Travis builds

fionawhim:
- Only pre-builds modules since apps will get built as part of tests
 - Removes `tsc --noEmit` check from module tests since building the
   module (done in `prepare`) runs `tsc` in order to output the types
 - Increases concurrency
 - Fixes app running check timeout for testcafe tests


jessicamarcus (approved):
lgtm 😁

PR #419, opened 2019-04-03
Adds icons to Access Boston

fionawhim:
Also updates fixture for new apps and renames career center icon since
that app was renamed.


paylakatel (approved):
lgtm!

PR #421, opened 2019-04-09
Strips non-digit characters from phone numbers

fionawhim:
Fix for Salesforce issue #36. We need to normalize at +1XXXXXXXXXX


jessicamarcus (approved):
lgtm!

PR #423, opened 2019-04-11
Shows / hides links based on city agency

fionawhim:
Also includes new way of invalidating login sessions so that when this
deploys we can ensure that people get their sessions populated
correctly.

See CityOfBoston/iam#565

### DIFF
--- a/services-js/access-boston/src/server/services/SamlAuth.ts
+++ b/services-js/access-boston/src/server/services/SamlAuth.ts
@@ -32,6 +32,7 @@ interface SamlAuthAssertion {
       userAccessToken?: string[];
       userMFARegistrationDate?: string[];
       isUserRegistered?: string[];
+      cobUserAgency?: string[];
     };
   };
 }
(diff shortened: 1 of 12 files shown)


services-js/access-boston/src/server/services/SamlAuth.ts
```diff
@@ -32,6 +32,7 @@ interface SamlAuthAssertion {
       userAccessToken?: string[];
       userMFARegistrationDate?: string[];
       isUserRegistered?: string[];
+      cobUserAgency?: string[];
```
jessicamarcus, line 35:
```suggestion
      userCobAgency?: string[];
```

jessicamarcus: ^^ just to be consistent with the existing naming (`userAccessToken` et al.)

jessicamarcus (changes requested):

services-js/access-boston/src/server/services/SamlAuth.ts
fionawhim: Those names come from the IAM team’s setup, so they’re not in our control.

jessicamarcus: Ah - bummer!

jessicamarcus (approved):

PR #425, opened 2019-04-16
Uses moment-timezone for correct MFA deadline

fionawhim:
Keeps us from saying the deadline is one day early because of GMT date
relative to Eastern date.


jessicamarcus (approved):
👍

PR #426, opened 2019-04-16
Storybook upgrade

jessicamarcus:
- Adds `storybook-common` module to (hopefully) simplify and keep consistency across projects
- Upgrades to Storybook 5 across all projects
- Adds viewport addon
- Reorganizes `react-fleet` Stories
- Updates dark Optimistic Blue value


modules-js/react-fleet/.storybook/addons.js
```diff
@@ -1 +1 @@
-require('@storybook/addon-actions/register');
+require('@cityofboston/storybook-common/addons.js');
```
fionawhim: Shouldn’t need the `js` here

modules-js/react-fleet/.storybook/manager-head.html
```diff
@@ -0,0 +1,10 @@
+<!--
+    Fleet’s css prevents the preview grid overlay from showing; since the grid
```
fionawhim, line 2: 👍

modules-js/storybook-common/.storybook/addons.js
```diff
@@ -0,0 +1 @@
+require('@cityofboston/storybook-common/addons.js');
```
fionawhim: Here, too (and below)

modules-js/storybook-common/.storybook/config.js
```diff
@@ -0,0 +1,11 @@
+import { addParameters, configure } from '@storybook/react';
+
+import { loadStories, storybookOptions } from '@cityofboston/storybook-common';
```
fionawhim, line 3: I think this should just be a relative import

modules-js/storybook-common/.storybook/config.js
```diff
@@ -0,0 +1,11 @@
...
+
+import './addons';
+
+const req = require.context('../src', true, /\.stories\.(jsx?|tsx?)$/);
+
+addParameters(storybookOptions(''));
```
fionawhim: Maybe make the title ”storybook-common Example”?

modules-js/storybook-common/src/storybook-common.ts
```diff
@@ -0,0 +1,46 @@
...
+    test: /\.html$/,
+    use: 'raw-loader',
+  });
+
+  config.resolve.extensions.push('.ts', '.tsx');
+
+  return config;
+}
+
+// Most projects will define req in their `config.js` like this:
```
fionawhim: Move this to a `/** */` JSDoc comment on `loadStories` so that it appears in IDEs?

modules-js/storybook-common/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "./node_modules/@cityofboston/config-typescript/tsconfig.default.json",
+  "compilerOptions": {
+    "rootDir": "src",
+    "outDir": "build",
+    "noUnusedParameters": false
```
fionawhim: Don’t add this here.

If you have an unused parameter, you can prefix it with `_` and TS will let it by.

Custom per-project compiler configs will only surprise / confuse us in future.

services-js/access-boston/.storybook/webpack.config.js
```diff
@@ -1,23 +1,3 @@
...
-  config.module.rules.push({
-    test: /\.html|\.yaml$/,
-    use: 'raw-loader',
-  });
-
-  config.resolve.extensions.push('.ts', '.tsx');
-
-  return config;
-};
+module.exports = ({ config }) => webpackConfig(config);
```
fionawhim, line 3: I love seeing all the red in this diff!

yarn.lock
```diff
@@ -51,6 +51,46 @@
...
+    "@babel/types" "^7.2.2"
+    convert-source-map "^1.1.0"
+    debug "^4.1.0"
+    json5 "^2.1.0"
+    lodash "^4.17.10"
+    resolve "^1.3.2"
+    semver "^5.4.1"
+    source-map "^0.5.0"
+
+"@babel/core@^7.4.3":
```
fionawhim: UGH. I'm not super into adding 2 new `@babel/core` versions, but I don't think there’s anything to be done. Next.js very much locks us to a specific version.

yarn.lock
```diff
@@ -100,6 +140,27 @@
     "@babel/traverse" "^7.1.0"
     "@babel/types" "^7.0.0"
 
+"@babel/helper-call-delegate@^7.4.0":
```
fionawhim, line 130: It looks like there are `^7.1.0` versions of a lot of these `^7.4.0` entries, which can be safely (or “safely”) upgraded to the 7.4.0.

Rather than perform surgery on yarn.lock you can straight delete the entries for "^7.1.0" dependencies and re-run `yarn install` and it’ll update those to `7.4.0` (unless `7.4.1` has snuck out in the interim).

modules-js/storybook-common/.storybook/config.js
jessicamarcus: Good call!

fionawhim (approved):
YAAAAYYYYYY!!!!

PR #429, opened 2019-04-18
Makes a local npm package for deploy-tools

fionawhim:
Lets us use the deploy tools from CI without installing all dependencies
and building, which will make it easier to parallelize test runs in
Travis, as we’d no longer be relying on the tests preparing the
deploy-tools package.


jessicamarcus (approved):
💯

jessicamarcus (approved):
Awesome wins in here! 😁

PR #430, opened 2019-04-23
yarn.lock updates to reduce total dependency count

fionawhim:
Also fixes deploy-tools’s report-updated-services for the new package
names and being run via npx.


jessicamarcus (approved):
💯

PR #431, opened 2019-04-23
Makes key generation password longer

fionawhim:
Fixes "You must type in 4 to 1023 characters" error when generating a
password.


jessicamarcus (approved):
👍

PR #433, opened 2019-04-24
Small changes to speed up Travis builds

fionawhim:
- Switches to xenial so we don't have to wait for as many services to
   start up
 - Caches npm and moves the environment variable setting into a
   check-for-changes.sh script so it can happen after lerna is installed
   globally (from the npm cache)


mmcgowanBOS (approved):
lgtm.

PR #436, opened 2019-05-08
Upgrades to Emotion 10

fionawhim:
We’re using the old-style "css" function, so no need to add the Babel
plugin right now to support the "css" prop in JSX.

Fixes Emotion hydration by putting it in the raw file rather than App
constructor. This prevents double-insertion of styles.

Also fixes the public notices CSS that was pushing the bottom bar off
the screen.


modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
```diff
@@ -511,10 +511,10 @@ const FIELDS_CONTAINER_STYLING = css({
     },
   },
   input: {
-    '-moz-appearance': 'textfield',
+    MozAppearance: 'textfield',
```
jessicamarcus, line 514: Whoa, didn’t realize vendor prefixes were available in Pascal case here.

jessicamarcus (approved):
👍

modules-js/react-fleet/src/form-elements/inputs/MemorableDateInput.tsx
fionawhim: Yah, there’s a warning now that kebab-case isn’t supported in objects.

PR #437, opened 2019-05-09
Migrates react-fleet to Emotion css prop

fionawhim:
Makes react-fleet just use the @emotion/core library and the css prop so
that it’s compatible with other apps switching to the css prop for
automatic SSR behavior.

 - Adds an eslint rule that enforces using the Emotion JSX processor
   when the css prop is used.
 - Installs jest-emotion for nicer CSS in our Storyshot diffs.
 - Updates usages of VISUALLY_HIDDEN and CLEAR_DEFAULT_STYLING since
   they’re now Emotion style objects, not class name strings.

Adds .babelrc to react-fleet’s .storybook so that we get our custom
Babel behavior, not Storybook’s default.

Upgrades @babel/preset-typescript to a version that correctly handles
not eliminating the "dead" import of a JSX factory.

Also removes Emotion from form-common, since it doesn’t have any UI
elements (and probably shouldn’t).

### DIFF
--- a/modules-js/react-fleet/src/form-elements/UploadPhoto.tsx
+++ b/modules-js/react-fleet/src/form-elements/UploadPhoto.tsx
@@ -1,7 +1,8 @@
+/** @jsx jsx */
 import React from 'react';
 import Dropzone from 'react-dropzone';
 
-import { css } from 'emotion';
+import { css, jsx } from '@emotion/core';
 
 import {
   FREEDOM_RED_DARK,
@@ -162,14 +163,15 @@ export default class UploadPhoto extends React.Component<Props, State> {
           onDrop={this.onDrop}
         >
           {({ getRootProps, getInputProps, isDragActive }) => (
-            <div {...getRootProps()} className={INPUT_CONTAINER_FOCUS_STYLING}>
+            <div {...getRootProps()} css={INPUT_CONTAINER_FOCUS_STYLING}>
               <input {...getInputProps()} />
 
               <div
                 ref={this.previewRef}
-                className={`${PREVIEW_CONTAINER_STYLING} ${
-                  isDragActive ? DRAG_RING_STYLING : ''
-                }`}
+                css={[
+                  PREVIEW_CONTAINER_STYLING,
+                  isDragActive && DRAG_RING_STYLING,
+                ]}
               >
                 {/* We keep the background element in the flow to preserve height / width */}
                 <div style={{ visibility: previewUrl ? 'hidden' : 'visible' }}>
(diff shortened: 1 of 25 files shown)


modules-js/react-fleet/src/form-elements/UploadPhoto.tsx
```diff
@@ -162,14 +163,15 @@ export default class UploadPhoto extends React.Component<Props, State> {
...
-            <div {...getRootProps()} className={INPUT_CONTAINER_FOCUS_STYLING}>
+            <div {...getRootProps()} css={INPUT_CONTAINER_FOCUS_STYLING}>
               <input {...getInputProps()} />
 
               <div
                 ref={this.previewRef}
-                className={`${PREVIEW_CONTAINER_STYLING} ${
-                  isDragActive ? DRAG_RING_STYLING : ''
-                }`}
+                css={[
```
jessicamarcus, line 171: Yay this is so much nicer!

services-js/registry-certs/client/birth/components/ApostilleRequestInstructions.tsx
```diff
@@ -1,6 +1,9 @@
+/** @jsx jsx */
+
 import React, { useEffect, useState } from 'react';
 
 import { css } from 'emotion';
```
jessicamarcus, line 5: Couldn’t this be

```suggestion
import { css, jsx } from '@emotion/core';
```
instead of separate lines?

services-js/registry-certs/client/birth/components/SupportingDocumentsInput.tsx
```diff
@@ -1,4 +1,6 @@
+/** @jsx jsx */
 import React from 'react';
+import { jsx } from '@emotion/core';
```
jessicamarcus, line 3: See above comment

services-js/registry-certs/client/common/CostSummary.tsx
```diff
@@ -1,7 +1,10 @@
+/** @jsx jsx */
+
 import React from 'react';
 import { css } from 'emotion';
```
jessicamarcus, line 4: See above

jessicamarcus (changes requested):

services-js/registry-certs/client/birth/components/ApostilleRequestInstructions.tsx
fionawhim: Not right now. `emotion`’s `css` is the legacy mode that generates a class name. `@emotion/core`’s `css` makes a style object.

I’m leaving these other apps in legacy mode for now.

jessicamarcus: Ah, got it! We can update as we go.

jessicamarcus (approved):

PR #438, opened 2019-05-09
Eliminates warnings in Jest runs

fionawhim:
- Guards against calling window.scroll when components mount
 - Changes file names in SupportingDocumentsInput to avoid key
   collisions
 - Delays Formik validation so it’s not running when the component
   unmounts
 - Upgrades velocity-react to the latest version that avoids warnings
 - Prevents a console.log in payment-webhooks
 - Mocks out window.scroll in RequestLayout tests


jessicamarcus (approved):
Nice dev experience improvements 👍

PR #440, opened 2019-05-10
Upgrades ESLint and switches to babel-eslint

fionawhim:
We’re standardizing on babel-eslint to have the same parser across all
of our files, and because typescript-eslint could fall out of date.

We have to use the beta of babel-eslint 11 because it (via the latest
@babel/core package) supports .babelrc files to know what plugins to
enable.

 - Switches from .eslintrc to .eslintrc.js so we can include helpful
   comments in the config file
 - Adds a few .babelrc files to packages that were missing them
 - Ran --fix on most source files, which picked up formating changes due
   to the Prettier upgrade

Also updates rxjs across the board to keep that version consistent,
which let us rely on inferred types in load-cases.ts.


jessicamarcus (approved):
lgtm!

PR #441, opened 2019-05-10
Fixes Ruby deploys to use packaged deploy-tools

fionawhim:
For some reason I left this change out of the one that established
deploy-tools.tgz.


jessicamarcus (approved):
👍

PR #442, opened 2019-05-13
Upgrades GraphQL libraries and tooling

fionawhim:
- Gets GraphQL to the 14.x release
 - Upgrades Apollo Server to 2.x, with the new "applyMiddleware" API
 - Switches from apollo-codegen to the apollo CLI tool
 - Updates eslint-apollo-plugin to the latest

The new version of apollo-server-* exports all of graphql-tools, so we
don’t need a separate import for makeExecutableSchema.

Restores the registry-certs server test in order to unit test persisted
queries.

Moves Rollbar error reporting out of the formatError callback and into a
GraphQL extension because 2.x Apollo Server doesn’t allow the whole
config to be generated per-request, just the context.

Uses "npx apollo" when generating types so that we don’t yarn install
the whole apollo CLI tool, which brings in a ton of other stuff (like
Heroku)

Also removes the unused "request documentation" API for registry-certs.

Note that Apollo Server 2.x uses GraphQL playground instead of GraphiQL.
It’s available just at /graphql. Unfortunately there’s no way to modify
its headers through ApolloServer, so it won’t work in staging / prod
where an API key is required.


jessicamarcus (approved):
👍

PR #448, opened 2019-05-15
Marriage Certs prep

jessicamarcus:
First sweep in preparation for Marriage Certs:

- moved and refactored Birth components that can be reused by Marriage into common directories
- refactored some constants
- updated Emotion usage across the project


services-js/registry-certs/client/PageWrapper.tsx
```diff
@@ -0,0 +1,67 @@
...
+        {footer}
+      </PageLayout>
+    );
+  }
+}
+
+// We put a hard top limit on the size, since with the responsive sizes on
+// sh-title the phrase “Request a birth certificate” gets too big and wraps
+// past 1245px wide. Since “marriage” contains more characters, the max font
+// size is dropped to 35px.
```
fionawhim, line 60: 👍

services-js/registry-certs/pages/_document.tsx
```diff
@@ -7,7 +7,7 @@ import { extractCritical } from 'emotion-server';
 import { StatusModal, ContactForm } from '@cityofboston/react-fleet';
```
fionawhim, line 6: Since you removed `emotion-server` from the `package.json` you’ll need to remove it from this file as well, and you can remove the `__NEXT_DATA__` shenanigans from here and `_app.tsx`

jessicamarcus: thanks for the catch!

services-js/registry-certs/pages/_document.tsx
```diff
@@ -47,14 +41,8 @@ export default class extends Document {
 
   constructor(props: Props) {
```
fionawhim, line 42: Can just delete this `constructor` now.

services-js/registry-certs/pages/_document.tsx
```diff
@@ -91,7 +78,7 @@ export default class extends Document {
                     client: {
                       javascript: {
                         source_map_enabled: true,
-                        code_version: "${rollbarVersion || buildId}",
+                        code_version: "${rollbarVersion}",
```
fionawhim: Keep the `buildId`

services-js/registry-certs/pages/_document.tsx
```diff
@@ -2,21 +2,17 @@
...
 
 type Props = {
-  __NEXT_DATA__: any;
   cacheParam: string;
   rollbarAccessToken: string | undefined;
   rollbarEnvironment: string;
   rollbarVersion: string | undefined;
-  // From Emotion’s hydration
-  ids?: string[];
   css: string;
```
fionawhim: remove `css` too

fionawhim (commented):
Few tweaks.

Also check both that there’s CSS in the HTML when you view source, and after the page loads if you inspect an element with Emotion styling there should be only one copy of the styles applied to it.

services-js/registry-certs/pages/_document.tsx
jessicamarcus: Hrm, my editor gives me a ts error without it. Will figure out what I’m missing.

services-js/registry-certs/pages/_document.tsx
jessicamarcus: The value was coming from `__NEXT_DATA__` - what should I assign it to instead?

services-js/registry-certs/pages/_document.tsx
fionawhim: yeah actually ok to leave it for now. Will be able to remove it when Next TypeScript types work right.

services-js/registry-certs/pages/_document.tsx
fionawhim: Keep using `__NEXT_DATA__`. It will still be on the `this.props` that `Document` has.

fionawhim:
Errors in Percy that need to be checked

fionawhim (approved):
LGTM

fionawhim:
If the Percy stuff all seems fine (mostly moves / renames?) then GTG for merge.

PR #453, opened 2019-05-17
Next canary prep

fionawhim:
An upcoming version of Next.js will have native TypeScript types. This
change adjusts some of our Next.js usage to be more compatible with
those changes when they land. (It still works on Next 8.0, though.)

 - Changes the "query" context property’s values to include string[],
   and makes it explicit that they might be undefined. Adds a "getParam"
   helper to next-client-common that can convert those values to a
   string.
 - Corrects the usage of the "conf" key when creating Next app objects.
   This seems to have only have worked because we were mutating a
   required module.
 - Calls url.format on Router.push’s argument to turn it into a string.
   Router.push did this before, but it looks like it won’t be part of
   the public API.
 - Changes Router event listening to use on/off instead of
   specially-named properties, since they won’t be in the public API.
 - Removes the unused ExtendedIncomingMessage type and adding "payload"
   to it.
 - Adds some defaults to tsconfig.default.json that the canaries are
   requiring be specified on tsconfig.json files. This introduces
   skipLibCheck, which may speed up compilation by not type-checking the
   (typically) auto-generated .d.ts files from packages.

### DIFF
--- a/modules-js/config-typescript/tsconfig.default.json
+++ b/modules-js/config-typescript/tsconfig.default.json
@@ -15,6 +15,8 @@
     "noImplicitThis": false,
     "resolveJsonModule": true,
     "experimentalDecorators": true,
+    "forceConsistentCasingInFileNames": false,
+    "skipLibCheck": true,
     "lib": [
       "dom",
       "es2017",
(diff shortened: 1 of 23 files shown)


jessicamarcus (approved):
Thanks for doing this prepwork, it will make life so much easier!

PR #457, opened 2019-05-29
Fire permit support for Permit Finder

fionawhim:
Re-works milestone step rendering to have less boilerplate for each
separate step, and instead calculate things from the current step / num
steps / step integers.

Adds in autolinking because some Fire contact info strings include URLs.
Uses this to link phone numbers as well, and keep them from wrapping.

Also:
 - Adds redirect from old PHP URLs
 - Removes unnecessary key logging
 - Removes contact person from GraphQL because it’s not rendered
   anywhere


jessicamarcus (approved):
👍

PR #458, opened 2019-05-30
Tracking branch for Next.js’s upcoming TypeScript

fionawhim:
The current 8.1.1 canary track includes native TypeScript support, which
includes official types for Next.js objects and out-of-the-box
TypeScript compilation. This is good! But it means we have to adapt some
of our existing TypeScript stuff.

As of this writing, Next.js enforces that the app directory contain its
own tsconfig.json file (zeit/next.js#7333), so this PR adds those as
necessary, inheriting from the parent package’s configuration. It also
requires a handful of specific configurations, so those are included as
well (zeit/next.js#7361) in those files.

This change also hoists acorn and acorn-dynamic-import because those two
packages need to be in a consistent place relative to each other and
webpack. (zeit/next.js#6240)

Currently the 'next' module does not export a type for its constructor
function, so we switch 'import next' to 'const next = require' to avoid
the typechecking.

We add "export {}" to the bottom of files that don’t do their own import
/ exporting to avoid an error due to the required "isolatedModules"
setting. (These can be removed if "isolatedModules" is removed.)

The bulk of the changes are updating _app and _document classes to the
nice new types.


fionawhim:
OK, this is the Next.js TypeScript update branch. It seems workingish as of the latest canary.

Unfortunately due to zeit/next.js#7361 our usages of `ts-node` may need to either change (maybe to `babel-node`?) or have their own `tsconfig.json` file that specifies a "commonjs" module type.

To update this to the latest canary, I think the easiest thing to do is delete the "next" entry from `yarn.lock` and then `yarn install`. You should also then comb through the `yarn.lock` and perhaps consolidate `@babel/…` entries.

jessicamarcus (approved):
👍

PR #497, opened 2019-08-13
Improves user experience, error reporting

jessicamarcus:
Currently, error state is only cleared from the relevant field onBlur, which can be confusing to users. This behavior has been changed to instead provide immediate feedback to the user.

I also changed a confusing component name, and moved some utility CSS out to the `react-fleet` module.

### DIFF
--- a/services-js/registry-certs/client/common/DateRangePicker.tsx
+++ b/services-js/registry-certs/client/common/DateRangePicker.tsx
@@ -4,7 +4,11 @@ import { css, jsx } from '@emotion/core';
 
 import { ChangeEvent, useEffect, useState } from 'react';
 
-import { FREEDOM_RED_LIGHT, MEDIA_MEDIUM } from '@cityofboston/react-fleet';
+import {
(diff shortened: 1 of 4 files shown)


services-js/registry-certs/client/common/DateRangePicker.tsx
```diff
@@ -137,38 +146,42 @@ export function DateSet(props: DateSetProps) {
...
-      setErrorMessage('Please enter a valid year.');
-      setYearInvalid(true);
-    } else if (+year > CURRENT_YEAR) {
-      setErrorMessage('The year cannot be in the future.');
-      setYearInvalid(true);
-    } else if (+year < LIMIT_YEAR) {
-      setErrorMessage(`Our records only go back to ${LIMIT_YEAR}.`);
-      setYearInvalid(true);
-    } else {
+  function isYearValid(): void {
```
phillipbentonkelly, line 160: Liked the streamlining ... 🗡

phillipbentonkelly (approved):

PR #501, opened 2019-08-22
Changes contact email address

jessicamarcus:
Patty has confirmed that the new email address is good to go. 😄


phillipbentonkelly (approved):
Looking good, approved percy

PR #512, opened 2019-09-25
Upgrades Next.js from 8.x.x to 9.x.x

jessicamarcus:
This PR is based on https://github.com/CityOfBoston/digital/pull/458/ and upgrades Next.js to latest across all apps **except** for 311; after upgrading (along with Babel from 7.1.x to 7.6.x), running Jest throws

```
[mobx] Encountered an uncaught exception that was thrown by a reaction or observer component, in: 'Reaction[RecentRequestRow#3350.render()] { Invariant Violation: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
```

Since 311 is not currently in production and it’s unknown whether it ever _will_ be, it was decided to leave it as-is for the time being rather than spend more time troubleshooting.

### DIFF
--- a/services-js/access-boston/src/server/access-boston.ts
+++ b/services-js/access-boston/src/server/access-boston.ts
@@ -11,7 +11,7 @@ import yar from 'yar';
 import cleanup from 'node-cleanup';
 import acceptLanguagePlugin from 'hapi-accept-language2';
 import hapiDevErrors from 'hapi-dev-errors';
-import next from 'next';
+const next = require('next');
 import { ApolloServer } from 'apollo-server-hapi';
 
 import { parse, Compile } from 'velocityjs';
--- a/modules-js/config-babel/browser.js
+++ b/modules-js/config-babel/browser.js
@@ -17,6 +17,7 @@ module.exports = api => ({
       require('@babel/preset-env'),
       {
         useBuiltIns: 'usage',
+        corejs: 2,
         modules: api.env('esm') ? false : undefined,
       },
     ],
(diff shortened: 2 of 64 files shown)


services-js/access-boston/src/server/access-boston.ts
```diff
@@ -11,7 +11,7 @@ import yar from 'yar';
 import cleanup from 'node-cleanup';
 import acceptLanguagePlugin from 'hapi-accept-language2';
 import hapiDevErrors from 'hapi-dev-errors';
-import next from 'next';
+const next = require('next');
```
phillipbentonkelly, line 14: Hadn't seen require and import mixed in the same file before. Is this required for the new version of Next, I see the same change in another file.

jessicamarcus: You know, I was following Fiona’s lead with the PR she made in preparation for this, but I didn’t actually test whether this syntax is necessary anymore... I’ll do so and update the PR.

phillipbentonkelly (approved):
Looks good, left a question.

services-js/access-boston/src/server/access-boston.ts
phillipbentonkelly: Ok, no prob. I don't see it as it blocker if it's working. I would guess, if turning `require`  into an `import` still works then I'd chuck-it to her just forgetting/copy-pasting. I'd done the same plenty of times lol. Just had the Q 💃

Issue #222, opened 2018-12-27
Update commissions search app to work with D8


zweishar:
The commissions search app is designed to ingest settings that are created by Drupal. In Drupal 7, those settings lived at `window.Drupal.settings.bos_commissions_search.bos_commissions_search_graphql_endpoint`
`window.Drupal.settings.bos_commissions_search.bos_commissions_search_graphql_api_key`

That path is different in D8. It is now:     `window.drupalSettings.bos_commissions.bos_commissions_search.graphql_endpoint`
`window.drupalSettings.bos_commissions.bos_commissions_search.api_key`

Ideally we could push an update to this app that would allow it to function in both D7 + D8.

I assigned @jessicamarcus and @mmcgowanBOS as I think you're best suited to take this over.

reillyzlab:
Needs to be looked into to see if this needs to happen.

Particularly if it's going to break the search.

jimafisk:
Example: https://www.boston.gov/civic-engagement/boards-and-commissions

reillyzlab:
Resolved so closing

PR #513, opened 2019-09-26
Handles correct path for either D7 or D8

jessicamarcus:
Adds a function to infer the version of Drupal and use the correct path for its settings in the commissions-search app, as described in https://github.com/CityOfBoston/digital/issues/222

**todo/note:** this change was committed with `--no-verify`, as there seems to be an issue with eslint not understanding jsx in this one case; I presume this has to do with the fact that this app was created with create-react-app and has a very different build than the rest of our webapps. Resolving this now is proving to be a time sink, but ought to be revisited in the near future.


phillipbentonkelly (approved):
Looks good.

PR #519, opened 2019-10-08
webapp build fixes

jessicamarcus:
- adds npm config to `Dockerfile`s
- adds separate config for `ts-node`

These additions should resolve the issues that had cropped up in some of our webapps after upgrading Next.js/Babel.


phillipbentonkelly (approved):
Looks Good!

PR #598, opened 2020-06-16
Registry-Certs: Hardcoded Outtage Message above Registry Apps content

phillipbentonkelly:
Tried making it a proper react component but that was a nightmare so I just copied the markup into ... ain't nobody got time for that smh


subaha-cob (dismissed):
That will work!

PR #739, opened 2021-01-21
Registry: Footer should be left aligned - all apps

phillipbentonkelly:
On all the registry applications, all of the footer content should be left aligned.

It's what reads, 'questions...'


subaha-cob (approved):
LGTM!

Issue #774, opened 2021-05-04
Update language on birth certificates to clarify middle name not needed 


reillyzlab:
Message/request from Patty:

On Step 3 of the online birth request - we [Registry Office] do not ask for the middle name because we do not need it and often it is not actually their middle name. We tend to get a couple of calls and a few emails a day asking if the middle name will be on the birth certificate - especially with newborns.

Can we edit it from:

What was their name at birth?
If they changed their name at some point, please use the name they were given at birth. If they were adopted, use their post-adoption name.

Edit it to:

What was their name at birth?
If they changed their name at some point, please use the name they were given at birth. If they were adopted, use their post-adoption name. Please note, we do not need their middle name to process your request. 

Screen shot of page this is on: ![image](https://user-images.githubusercontent.com/20731897/117030256-42d0c100-acb4-11eb-8a77-10c075cb7d18.png)

jimmurphy-boston:
Patty reviewed staging and is good to go.

PR #780, opened 2021-06-10
Registry-Certs: Update language on birth certificates

phillipbentonkelly:
[Registry-Certs: Update language on birth certificates to clarify middle name is not needed](https://github.com/CityOfBoston/digital/issues/774)


mmcgowanBOS (approved):
LGTM.

PR #783, opened 2021-06-30
MS-SQL/Registry option flag

phillipbentonkelly:
Enabling encryption flag on `ms-sql` to be set as an option, removed un-needed coded


subaha-cob (approved):
LGTM!

PR #784, opened 2021-07-02
Registry-Certs(Marriage Intention): Fix value marriage num value

phillipbentonkelly:
Normalize default option for last marriage status


mmcgowanBOS (approved):
LGTM.

PR #796, opened 2021-08-10
Registry-Certs: Use multiSubnetFailover on DB connection

phillipbentonkelly:
As part of the DB migration, we need to start connecting to clusters instead of DB servers. This is an attempt to connect using `multiSubnetFailover` from the cluster so its connection bounces until it finds a source or all sources have been checked.


subaha-cob (dismissed):
LGTM!

mmcgowanBOS (approved):
LGTM.

PR #810, opened 2021-10-21
Access-Boston: Notice/Alert Header (Optional)

phillipbentonkelly:
Allow a message (`Notice`) to be displayed above each app group on the Access-Boston Dashboard by adding a `notice: ....` node to the config section of that app group.


subaha-cob (dismissed):
LGTM

PR #830, opened 2021-12-16
Access-Boston: Setting Notice text from Config

phillipbentonkelly:
Setting up banner message area (`Notice`) to be populated via configs. Previously, this was hardcoded, now whether the `notice` is shown can be set by including a `notice` node in the config file for the [environments](https://github.com/CityOfBoston/access-boston-config/tree/master/src/configs). If the node below is missing from the top level of the config file then the `notice` header will not be shown.

```yaml
notice:
  label: 'Notice'
  pretext: 'Beacon, the new HR self-service portal, is now live!'
  text: 'Look for the lighthouse icon below.'
```


subaha-cob (approved):
LGTM!

PR #973, opened 2023-05-02
DIG-2171: Remove Gender Question from Marriage Intention Form

phillipbentonkelly:
Marriage Intention Form Update:
- DIG-2171: Remove Gender Question from Marriage Intention Form
- DIG-2105: Add survey to marriage intention form application
- DIG-2286: User cannot pick IV or V suffix on Marriage Intention form

Due to the deploy process being broken at the moment we implement an AD-HOC method to deploy to Staging and PROD through Dockerfile and Compose. Process is outline in the Dockerfile


subaha-cob (approved):
I tried to really look at it :) I think it looks good!

PR #1029, opened 2024-11-04
DIG-5324: Success Screen

xuhongkang:


services-js/access-boston/src/server/services/SamlAuthFake.ts
xuhongkang, line 1: Should this change be persistent?

xuhongkang (commented):
All current changes should match the prototype UX design in PC view, but not in mobile view. Has the mobile view UX been implemented?

phillipbentonkelly (approved):
LGTM

